PR #40, opened 2025-12-06
Add authentication and remote access support

marcelveldt:
I asked Claude to add the support for authentication and remote acces.
Feel free to edit or just use as reference and remove this PR and branch.  

It adds support for all 3 login methods:

- Let the user enter a long lived token manuallly
- Let the user enter username+password
- Use the login oAuth flow with callback (also allows "login with HA")

maybe just only pick the oAuth login flow with the callback ?

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -43,10 +43,14 @@ kotlin {
             implementation(libs.koin.androidx.compose)
 
             implementation(libs.androidx.media)
+            implementation(libs.androidx.browser)
 
             implementation(libs.coil)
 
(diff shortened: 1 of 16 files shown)


marcelveldt:
This PR is just here for reference for @formatBCE to have an idea of what is globally needed. It's probably full of AI hallucinations as always with AI coding.

The app is still in very early stages so it will take some time to implement the required changes.

If you depend on the app, just not yet update to MA 2.7 or use the web app in the meanwhile.

formatBCE:
I promise I will tackle this asap. :)

PR #78, opened 2026-02-05
Sendspin architecture refactoring

formatBCE:


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/PlayersTopBar.kt
```diff
@@ -34,9 +36,11 @@ fun PlayersTopBar(
         verticalAlignment = Alignment.CenterVertically
     ) {
         IconButton(
+            modifier = Modifier.size(32.dp),
```
seadowg: @formatBCE just noticed these had dropped in size again - these changes lower the button sizes below the minimum touch target size (https://developer.android.com/develop/ui/compose/accessibility/api-defaults).

Was this on purpose? It seems unrelated to the other changes here. Would be interested to understand the rationale if so.

formatBCE: It just looks better. :) there's too much space.

PR #79, opened 2026-02-06
Add lint and test checks for PRs

seadowg:
As promised in #76.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
@@ -1,6 +1,10 @@
 package io.music_assistant.client.utils
 
 import androidx.compose.ui.Modifier
+import io.ktor.http.URLBuilder
+import io.ktor.http.URLProtocol
+import io.ktor.network.sockets.InetSocketAddress
+import io.ktor.util.network.NetworkAddress
 import io.music_assistant.client.api.Answer
 import kotlin.time.Duration
 import kotlin.time.DurationUnit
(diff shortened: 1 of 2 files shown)


formatBCE:
> @formatBCE [`ExtTest#testInvalidHosts`](https://github.com/music-assistant/mobile-app/blob/93240c97a5f2a55624a924f21d3fd395c673b438/composeApp/src/commonTest/kotlin/io/music_assistant/client/utils/ExtTest.kt#L33) is failing. My guess (from looking at the code/tests) is that the code is in the wrong here, but I wanted to confirm that with you before making any changes. 

Yeah everything there seems reasonable testing. Canfix the field?

composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
```diff
@@ -24,6 +28,7 @@ fun Duration?.formatDuration() =
     } ?: "--:--"
 
 fun String.isValidHost(): Boolean {
+    val ipLikePattern = Regex("^(-?\\d{1,3}\\.)+-?\\d{1,3}$")
```
seadowg: The trick here to get both the IP and hostname cases working is to loosely detect IP addresses (any string that's sections of up to 3-digit numbers separated by dots) and then use the appropriate matcher to validate. Otherwise, pretty much any invalid IP will be (syntactically) valid as a hostname.

composeApp/lint-baseline.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
seadowg: These will want fixed eventually, but I didn't think that it made sense to dive into Android Auto stuff just to get PR checks working. Maybe something to file an issue around?

PR #93, opened 2026-02-18
Don't focus search input when there's a query

seadowg:
# Search changes

This prevents focus from being grabbed by the search input when there's already a query. That stops situations like this when pressing back from navigating to an album/artist from results:

Screenshot_20260218_090005

Long term there's potentially a more involved solution for keeping track of focus state in something like a ViewModel. I lived with this change for a couple of days though, and it felt pretty much right to me.

# Testing

The biggest change here is bootstrapping the ability to write tests for the multiplatform Composables in `commonMain`. The actual changes and tests added here are very small! 

Testing for Compose Multiplatform projects is [still pretty experimental](https://kotlinlang.org/docs/multiplatform/compose-test.html). From my perspective, assuming we want to easily run tests in CI without an Android emulator or iOS simulator, there's basically two options:

1. Host all common Compose tests in `androidUnitTest` and run them with Robolectric (using the Compose/Robolectric testing dependencies). This allows running tests in IDEs and on the command line/CLI (`./gradlew testDebug`), but it means that there's a risk of introducing Android specific passes. If all the Compose code is hosted in `commonMain` that's hopefully minimized. One perhaps less obvious advantage of this approach is that Compose testing frameworks are much more mature (and exposed) in Android than anywhere else - there's likely stability that comes with that.
2. Use `commonTest` (with the experimental test APIs) and utilize the desktop target for running tests. This again makes it easy to run tests in the IDE and also gives you a way to run them on the command line/CLI (`./gradlew jvmTest`), but has the disadvantage of managing the extra target if you're not actually using desktop. In the future, Robolectric support could come to `commonTest` and eliminate that drawback though. This also gives you the advantage of being able to run tests on an iOS simulator whenever you want.

I've included option 1 here which I think I prefer for the moment, but I also played with 2. The latter doesn't work out of the box as it seems we need to update some of the Compose Multiplatform dependencies and plugins to make them compatible with the current testing libraries. That said, I do think it could work, and it wasn't that hard to write quick `actual` stubs for the desktop target.

### DIFF
--- a/composeApp/src/androidUnitTest/kotlin/io/music_assistant/client/ui/compose/search/SearchInputTest.kt
+++ b/composeApp/src/androidUnitTest/kotlin/io/music_assistant/client/ui/compose/search/SearchInputTest.kt
@@ -1,6 +1,7 @@
 package io.music_assistant.client.ui.compose.search
 
 import androidx.compose.ui.test.assertIsFocused
+import androidx.compose.ui.test.assertIsNotFocused
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -27,6 +27,7 @@ gen/
 
 # Local configuration file (sdk path, etc)
 local.properties
+.envrc
```
seadowg: [`direnv`](https://direnv.net/) is definitely useful for making sure you are using the correct JVM in your terminal. In generally have it set to the Android Studio one to keep things consistent:

```
export JAVA_HOME=/Applications/Android\ Studio.app/Contents/jbr/Contents/Home/
```

formatBCE (approved):
Love this!

PR #127, opened 2026-02-26
Use Split Button on item details screen

seadowg:
Happy to find out that there was a multiplatform implementation - we just needed to upgrade!

Screenshot_20260226_101957


formatBCE (approved):
OMG this looks awesome.

Issue #113, opened 2026-02-21
Landscape orientation has some issues.


AkoreJepade:
Portrait mode works great on phone but landscape mode on tablets not so much so.

formatBCE:
Agree. I tried several ideas with landscape, and they don't work. We desperately need UX guy. :)

AkoreJepade:
I nominate [https://github.com/chrisuthe](url) as his SendspinDroid works very well in both portrait and landscape. Unfortunately my programming dates back to COBOL, C++, Object oriented C++ and Basic. Heck when I first took electronics in High School, the first year (9th grade 1978) they were still teaching TUBES!

AkoreJepade:
![Image](https://github.com/user-attachments/assets/f8ab1951-fd2f-4f0c-80d2-a5939cf1256e)
![Image](https://github.com/user-attachments/assets/e23bd53a-9cf4-49e9-ace2-85800b969d59)
![Image](https://github.com/user-attachments/assets/603d3055-0377-41e8-a916-580dc1b65003)
![Image](https://github.com/user-attachments/assets/77eb6923-e8ee-4c78-8709-971f2ed143d8)

seadowg:
Ok so the main candidates to get reworks for larger screens based on those screenshots are:

- [x] Album/artist (item) screen
- [x] Home screen
- [x] Expanded player/queue

@formatBCE if you assign me onto this I can look at it next!

chrisuthe:
I'll let @seadowg  run let me know if I can help!

seadowg:
I'm thinking that for improving home in landscape and for generally improving navigation it'd be good to adopt the typical Material 3 navigation components with a [bottom nav bar](https://m3.material.io/components/navigation-bar/overview) for smaller devices and a nav rail on larger devices (https://m3.material.io/components/navigation-rail/overview). A lot of music players (I played around with Tidal, Spotify and Symfonium on Android) take this approach and either have the "player bar" (currently playing info/controls) as a [floating bar](https://m3.material.io/components/toolbars/overview) above (or at the bottom on larger devices) or as an extension of the bottom nav bar. I think the former makes more sense for us given that you can page through the different players so it'd be good to keep separate.

Here's a very dodgy mock up (of the compact experience) I made while playing around with ideas in the codebase:

<img width="270" height="1200" alt="Image" src="https://github.com/user-attachments/assets/cbb10204-b40b-48b6-a7ff-0a4e0e502912" />

[...]

PR #133, opened 2026-03-01
Finish off item page rework

### LINKED ISSUE #108: Refactoring item details page
## Problem

Currently, the version of an album isn't displayed anywhere (in search results, artist pages or album pages). This makes it pretty much impossible to find the right version in cases where you can't differentiate using artwork (or if the artwork differences are really subtle).

## Solution

Music Assistant desktop/web doesn't handle different versions of albums in search, but in the artist view versions appear as "\<album title\> (\<version\>)" and on album pages the version is shown under the title.

Ideally we'd address this problem in all three places by also using the "\<album title\> (\<version\>)" formatting in search results as well (and matching the behaviour on artist and album pages). One complication is that currently album pages show the title in the "app bar" so that might need rethought to accommodate more info.

seadowg:
Closes #108
Work towards #113

This takes care of the rest of the issue:

- Display a provider icon on the artwork
- Display version on album details screen
- Optimize the layout for landscape mode
- Move non-play actions to a top right overflow
- Make play and non-play actions language and icons consistent with desktop

I played around with different options for the item screen's "landscape" mode and settled on a variation derived from the desktop layout:

Screenshot_20260301_170246

This mode kicks in at the ["expanded" size class](https://developer.android.com/develop/ui/views/layout/use-window-size-classes) (used for the screenshot), with compact and medium using the more "portrait" oriented layout.


composeApp/src/androidUnitTest/kotlin/io/music_assistant/client/ui/compose/support/Scrollable.kt
```diff
@@ -0,0 +1,19 @@
+package io.music_assistant.client.ui.compose.support
+
+import androidx.compose.ui.test.SemanticsMatcher
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.junit4.ComposeTestRule
+import androidx.compose.ui.test.onFirst
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performScrollToNode
+
+fun ComposeTestRule.inScrollable(scrollableTag: String, block: ScrollableScope.() -> Unit) {
```
seadowg: I'm a little shocked there aren't more helpers in Compose's testing libraries to deal with scrolling out of the box.

composeApp/src/commonTest/kotlin/io/music_assistant/client/utils/support/Mocks.kt
```diff
@@ -0,0 +1,45 @@
+package io.music_assistant.client.utils.support
+
+abstract class MockFunction {
```
seadowg: I'd usually prefer just rolling my own fakes/mocks than using a framework, but Compose's callback heavy interfaces are an ideal candidate for the latter as you end up with a lot more anonymous function parameters than domain specific interfaces. I had a quick look at KMP safe options (as opposed to using a JVM specific standard like Mockito), and it honestly seemed easier to just quickly roll a utility for ourselves while we let that scene mature.

formatBCE (approved):
Thanks for the foundational steps!

PR #141, opened 2026-03-05
Fixed playback of OPUS on iOS

baileyboy0304:
Opus decode was producing "slow" audio which was traced to the audio buffer being processed incorrectly in AudioDecoders.swift


formatBCE (changes requested):
Something definitely went wrong. Looks like decompiled library code was added to Git. Please fix this. 443 files is great number, but too much for this fix. :)

PR #147, opened 2026-03-09
Add docs on structure

seadowg:
Just some small additions to the `README` to describe the current structure of the project.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -23,12 +23,13 @@ android {
         }
     }
     signingConfigs {
-        getByName("debug") {
+        create("selfSignedRelease") {
             storeFile = file(System.getProperty("user.home") + "/.android/debug.keystore")
             storePassword = "android"
             keyAlias = "androiddebugkey"
             keyPassword = "android"
         }
+
         create("release") {
             val props = Properties().apply {
                 val file = project.file("keystore.properties")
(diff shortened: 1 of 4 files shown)


androidApp/build.gradle.kts
```diff
@@ -23,12 +23,13 @@ android {
         }
     }
     signingConfigs {
-        getByName("debug") {
```
seadowg, line 26: This was causing failures for PRs without the keystore configured (https://github.com/music-assistant/mobile-app/actions/runs/22850043446/job/66276058579?pr=147).

I've used this trick on a few other projects: a separate "self signed" build variant that uses the debug keystore. Gives you the same benefits, but doesn't cause any problems for the default "debug" variant.

formatBCE: Will this build with stable signature?

seadowg: It's using the same signing config as before just renamed, so it should as long as the `.android/debug.keystore` has been configured for the action on `dev`. If you try it out locally it should use your machine's debug signature.

The PR checks are just using standard `debug` so won't have a stable signature, but we don't "export" that APK anyway as far as I'm aware.

formatBCE: I just saw you removed the step with restoring keystone from settings in check-pr. Thus ask.

seadowg: Yeah, the Gradle tooling should generate one for the `debug` variant in the PR check builds!

formatBCE: Makes sense.

PR #148, opened 2026-03-10
Unify player selection and grouping

### LINKED ISSUE #80: Refresh and Select Player actions are very small
I personally find these actions too small to reliably hit (especially when they're in the top corners):

Image

These are smaller than the recommended tap target sizes for Android and iOS ((the recommendations being [48dp on Android](https://support.google.com/accessibility/android/answer/7101858?hl=en-GB) and [44dp (points) on iOS](https://developer.apple.com/design/human-interface-guidelines/accessibility#Mobility).

As discussed with @formatBCE, there's probably a way to rearrange things so that these actions can be bigger without increasing the vertical space used or general businesses of the collapsed player bar. Two options that come to mind for me are:

1. Minimizing the number of actions on the collapsed player in portrait mode - track/album title (stacked) on the left and then player select and play/pause on the right. This would be similar to Roon.
2. Only show player selection/refresh when the player is fully expanded to allow more space for art. This would be more similar to Plexamp or Symfonium.
3. A combination of both (art and player selection) with marqueeing of long track/album titles like Spotify.

seadowg:
Closes #80
Work towards #113


seadowg:
@formatBCE this isn't finished yet, but I think it's a good point to discuss! You'll be able to pull this down and give it a go, but here are some screenshots:

Screenshot_20260310_084012

Screenshot_20260310_084036

Screenshot_20260310_084046

The idea here is to move selecting players and grouping players one flow while also reducing the vertical space used by the compact player bar. Clicking the player name in the expanded player opens the same flow. There's some more polish needed on the design (dismissing without a back button, dialog titles, visual tweaks etc.). I played with this flow for a couple of days and have liked it, but definitely have things to discuss:

- Should "transfer" live in the player selection as well?
- We could move the "group" action to be a right aligned button in the selected player (and transfer if we moved it also)
- Do we want radio buttons here or would icons for the players (the ones the user sets) be more useful? We'd probably need to use a different background for the "selected" player if we went with the latter. Showing both is potentially messy, but we could try it. 

Let me know what you think and if you have other thoughts!

EDIT: I've also noticed that there seems to be an off by one error where you sometimes get the wrong player when selecting one from the menu (as opposed to swiping). It looks like that's in `dev` as well, but I can look at that as part of this.

formatBCE:
Yeah sometimes the pager just can't cope with fast auto-scrolling, and ends on one before the actual selection. I think it's something to do with launched effects there... Will need to check it out.

On the main topic - whoosh, now it looks pretty hard to get into grouping. Also it doesn't seem to be natural place for that functionality. How do we explain to user that to get to grouping they need to open common players list?
On the other hand, with this narrow section I'm not sure where to put that otherwise too. Honestly, player name under the speaker icon is also sub-optimal. You, for one, have short player names - but I, for instance, have something like "Kids Bedroom Koala". That would mess your design right away.

Probably we should stop somewhere with that shrinking? Or just accept that we move most of functionality to the full player view?

seadowg:
> Yeah sometimes the pager just can't cope with fast auto-scrolling, and ends on one before the actual selection. I think it's something to do with launched effects there... Will need to check it out.

Yeah definitely feels like it could be related to a `LaunchedEffect`. As I say, I can take a look as part of this seeing as I'm in there anyway!

> On the main topic - whoosh, now it looks pretty hard to get into grouping. Also it doesn't seem to be natural place for that functionality. How do we explain to user that to get to grouping they need to open common players list?

I mean currently you click the "player" to get to grouping and that's still true right? It's an extra click, but the way in is the same. This is a similar mental model to the desktop experience where you click the current player to get to player "control" (selection, transfer, grouping etc.).

> Honestly, player name under the speaker icon is also sub-optimal. You, for one, have short player names - but I, for instance, have something like "Kids Bedroom Koala". That would mess your design right away.

Yeah agreed. I think it will be nice for wider screens. On compact screens we could hide the name and just show the icon (which we need to add support for)?

> On the other hand, with this narrow section I'm not sure where to put that otherwise too.

One compromise here if you really don't like the new player selection button on the compact screens (where showing full player names is going to be awkward) is just to keep the player name button for them, but have that display the unified selection/grouping dialog. Wider screens could use the style shown here. That gets us the vertical space wins in landscape. 

> Or just accept that we move most of functionality to the full player view?

I do personally like to have easy access to player selection/grouping. This app will most likely be the main remote control for MA for a lot of folks and I think keeping player control a "primary" action (like you already have it) is a good.

seadowg:
> One compromise here if you really don't like the new player selection button on the compact screens (where showing full player names is going to be awkward) is just to keep the player name button for them, but have that display the unified selection/grouping dialog. Wider screens could use the style shown here. That gets us the vertical space wins in landscape.

@formatBCE I decided to give this a try this and see how it feels. Here's the compact layout:

Screenshot_20260312_105621

...and here's in landscape on a tablet:

Screenshot_20260312_105645

Interested to see what you think about this as the general approach! As I said before, still a lot to tweak in the selection dialog (need to remove hide/disable group button for players without that option, tweak layout etc.).

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
```diff
@@ -110,10 +110,11 @@ fun HomeScreen(
...
             ?: return@LaunchedEffect
         val target = currentData.selectedPlayerIndex ?: return@LaunchedEffect
-        if (playerPagerState.currentPage != target) {
-            playerPagerState.scrollToPage(target)
+        if (!playerPagerState.isScrollInProgress) {
+            playerPagerState.animateScrollToPage(target)
         }
-        snapshotFlow { playerPagerState.currentPage }.collect { currentPage ->
+
+        snapshotFlow { playerPagerState.settledPage }.collect { currentPage ->
```
seadowg, line 117: > Yeah sometimes the pager just can't cope with fast auto-scrolling, and ends on one before the actual selection. I think it's something to do with launched effects there... Will need to check it out.


This change and the previous commit appear to fix this. I've got it on my device so I can test it over the next day or so to double-check.

formatBCE:
There's still no room for long player name in landscape mode. Maybe we arrange them side by side on the right? Or even use same graphical element (button with player name) instead of stacked icon+name?

seadowg:
> There's still no room for long player name in landscape mode.

It can grow pretty big:

Screenshot_20260312_201547

There's room beyond that (the whole right side is spare right now), but we can elipsize to stop it getting too out of hand. Are you seeing it looking more squished or do you just mean it doesn't look good with the text like that?

> Or even use same graphical element (button with player name) instead of stacked icon+name?

Yeah, we could just use a standard button if it feels like this doesn't work.

formatBCE:
Yeah, I'd prefer unified look for the control in both modes. Having it on the right though is great decision.

seadowg:
> Yeah, I'd prefer unified look for the control in both modes. Having it on the right though is great decision.

Nice! I like that direction - I'll give it a go.

seadowg:
> Nice! I like that direction - I'll give it a go.

@formatBCE done! I'm travelling this weekend and don't have access to my MA server to test this properly, but if you get a chance to take a look let me know if there are any other tweaks you'd like. As with anything, there are many more things we could do here, but it's probably at a good place to think about merging.

formatBCE:
While I hate additional clicks, I must admit that it makes UI cleaner. :)
Found trouble with the dialog though:

seadowg:
> Found trouble with the dialog though:

Ach yeah I could have sworn there was something else I'd meant to do: the list of players isn't scrollable so it'll cut off weirdly once it goes beyond the max dialog height. I'll fix that.

seadowg:
> Found trouble with the dialog though:

That should be fixed now. I noticed you had put a max height for the list of players on the group settings, so I've used the same one for the player list.

formatBCE:
Okay, looks like we're ready to merge this? Apart from the fact that I find this dialog pretty confusing, it works great :)

seadowg:
> Okay, looks like we're ready to merge this?

Yup!

> Apart from the fact that I find this dialog pretty confusing, it works great :)

Is there anything you'd like to tweak before merging? This flow basically makes sense to me because it's similar to how Roon and how desktop MA works, and I'm used to those products - I totally get it might not be as intuitive to others not coming from that context. An option I'd suggested earlier was to move the "Group" button into the player item? Going further, we could add a three dots overflow menu to the player item that contains group, transfer etc. like on desktop?

formatBCE (approved):

PR #158, opened 2026-03-26
Added ability to share regular and crash logs. Fixed genres crash.

formatBCE:


cociweb:
@formatBCE Great improvement! Could you please redact (at least partially) the personal infos such as server ip/dns,  auth info, tokens and client id, etc

formatBCE:
> @formatBCE Great improvement! Could you please redact (at least partially) the personal infos such as server ip/dns,  auth info, tokens and client id, etc

Yup, will do. This was done fast to get logs for particular issue. Will be cleaning up now.

PR #162, opened 2026-03-26
Sanitized logs from sensitive data.

formatBCE:


cociweb:
@formatBCE Thx, fast as a rocket! But I still see userinfo/userid and tokens mostly in Error/: State Authenticated() and in  Debug/SendspinWsHandler: Sending text message: {
    "type": ".."} where type is auth or client/hello

formatBCE:
> @formatBCE Thx, fast as a rocket! But I still see userinfo/userid and tokens mostly in Error/: State Authenticated() and in  Debug/SendspinWsHandler: Sending text message: {
>     "type": ".."} where type is auth or client/hello

Okay, apparently I haven't found all of them :)
Thanks, will check it!

PR #171, opened 2026-03-29
Tweak compact player spacing

seadowg:
Just some quick tweaks for the compact player/player controls to help in cases where the track name is longer than the available space.

Before:

Screenshot 2026-03-29 at 11 39 22

After:

Screenshot 2026-03-29 at 11 40 40

EDIT: while testing I also noticed a small problem with predictive back that [I've fixed](https://github.com/music-assistant/mobile-app/pull/171#discussion_r3006023990).

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
@@ -19,6 +19,7 @@ import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Surface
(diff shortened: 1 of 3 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -76,35 +77,37 @@ fun ItemDetailsScreen(
...
-        { modifier, provider ->
-            actionsViewModel.getProviderIcon(provider)
-                ?.let { ProviderIcon(modifier, it) }
-        },
-        viewModel::onPlayClick,
-        viewModel::toggleItemsRowMode,
-        viewModel::onChapterClick,
-        viewModel::onPlayClick
-    )
+    Surface {
```
seadowg, line 80: Just noticed this while playing around: when performing a back gesture on Android 16 the screen shows as transparent during the predictive back gesture as it's "surface" is provided by the `Scaffold` around `NavDisplay`.

formatBCE (approved):
Awesome!

Issue #179, opened 2026-03-30
feature: Support TLS 1.3 for the WS/Sendspin connections


schildbach:
I'm coming from #172. On my Pixel 7a with the latest (officially released) Android 16 and all apps up to date, the app (commit: 09ff43e7ec7bf9e2c18761bbe84710f22426ba8f) can't connect to a server (actually reverse proxy) via TLS 1.3.

I'm using traefik 3.6.12 and by default, I restrict TLS to 1.3 (and above). With this config, and with "Use secure connection" checked in the app, it app logs this error: "Received alert during handshake. Level: FATAL, code: ProtocolVersion". 

```
1774883987006 Info/MainDataSource: SessionState changed: NoServerData
1774883987007 Info/MainDataSource: Disconnected (NoServerData) - clearing data
1774883987008 Info/SendspinClientFactory: Destroying shared audio pipeline
1774883987008 Info/MainDataSource: Clearing all cached data
1774883987210 Info/ServiceClient: App foregrounded
1774884051465 Info/MainDataSource: SessionState changed: Connecting
1774884051465 Info/MainDataSource: Connecting - stopping Sendspin
1774884051466 Info/SendspinClientFactory: Destroying shared audio pipeline
1774884051778 Warn/DirectTransport: Initial connection failed: Received alert during handshake. Level: FATAL, code: ProtocolVersion
1774884051779 Info/MainDataSource: SessionState changed: Error
1774884051779 Warn/MainDataSource: Persistent error with no data to preserve
```

When relaxing the server/proxy to TLS 1.2 and above, the app can connect as expected.

formatBCE:
Should be fixed in next build.

basrieter:
Wad did indeed fixed? Downloaded the latest debug build and can't connect when enabling HTTPS in MA server.

formatBCE:
> Wad did indeed fixed? Downloaded the latest debug build and can't connect when enabling HTTPS in MA server.

Is it TLS 1.3 problem? Or just HTTPS connection in general?

My HTTPS MA server is perfectly accessible.

schildbach:
I just checked version `android-v0.3.0`.

The connection to Music Assistant 2.8.3 now works fine over TLS 1.3. Thanks for the fix!

[...]

PR #180, opened 2026-03-30
Split HTTP engines by platform. 

formatBCE:
Fixed crash for background service. Fixed local player init/deinit.


formatBCE:
Should fix #179

PR #184, opened 2026-03-31
Add Google Play API access test step

MarvinSchenkel:
Added a step to test Google Play API access with curl.


MarvinSchenkel:
(no need to merge)

formatBCE (approved):

PR #194, opened 2026-04-01
Host compact player in floating toolbar

seadowg:
Work towards #113

As discussed in the issue, this moves the "compact" player to a floating toolbar. This allows it to be distinct when a bottom nav bar is added.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
@@ -32,7 +32,7 @@ fun FloatingBar(modifier: Modifier = Modifier, onClick: () -> Unit = {}, content
 }
 
 object FloatingBarDefaults {
-    val padding = 4.dp
+    val padding = 8.dp
 }
 
 @Preview
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
@@ -22,13 +22,17 @@ fun FloatingBar(modifier: Modifier = Modifier, onClick: () -> Unit = {}, content
     Box(
         modifier
             .fillMaxWidth()
-            .padding(16.dp)
+            .padding(FloatingBarDefaults.padding)
             .clip(RoundedCornerShape(16.dp))
             .background(MaterialTheme.colorScheme.surfaceContainerHigh)
             .clickable { onClick() }
     ) {
         content()
     }
+}
+
+object FloatingBarDefaults {
+    val padding = 8.dp
 }
 
 @Preview
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
@@ -1,0 +1,57 @@
+package io.music_assistant.client.ui.compose.home
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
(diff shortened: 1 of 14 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun FloatingBar(modifier: Modifier = Modifier, onClick: () -> Unit = {}, content: @Composable () -> Unit) {
```
seadowg, line 21: I ended up creating a custom component for this instead of using [`HorizontalFloatingBar`](https://kotlinlang.org/api/compose-multiplatform/material3/androidx.compose.material3/-horizontal-floating-toolbar.html). I initially did go with Material 3 implementation, but I ran into a few weird bugs with sizing and full bar clicks. Given that that component is still experimental as well, I think it feels ok to use a custom one for now.

formatBCE:
Don't you think that this is unnecessary element? We're basically shrinking vertical space even more by introducing bottom unusable space... Sure it looks good - but honestly it annoys me on MASS web page - yet to see on Android.

Just asking. :)

formatBCE:
P.S. Installed and tried - IDK what to say. I Don't want to diminish your work for sure, it's awesome. 
Probably it's just personal preference.
But I really don't like it. It gets lost in main content, the underlying content makes it look weird.
At the minimum - I'd reduce margins twice. At the maximum - just don't use this element. :(

seadowg:
> P.S. Installed and tried - IDK what to say. I Don't want to diminish your work for sure, it's awesome.
Probably it's just personal preference.
But I really don't like it. It gets lost in main content, the underlying content makes it look weird.

All good - no offence taken. So my main reason for thinking we should do this was really just keeping it distinct from the nav bar. In a lot of music apps, the "player bar" and the nav bar are combined, but I don't think that's going to work for us because of the paging interaction.

I was thinking it would be nice to stage this, so the floating player comes in first, but I think you're right that the change feels strange in isolation. How about I include the most basic version of the nav bar (just home and settings) as well in this PR? I think that will give us a better idea of if floating makes sense or if we want to do something else.

formatBCE:
I'm already curious how bad will it be with top bar, player view and nav bar. We easily eat 50% of screen with that.
Let's try with bottom me u to see how bad it is. :)

seadowg:
> I'm already curious how bad will it be with top bar, player view and nav bar. We easily eat 50% of screen with that.

We'll be able to remove/reduce the top bar one nav is handled at the bottom I reckon!

> Let's try with bottom me u to see how bad it is. :)

Loving the enthusiastic pessimism 😂. Let's go!

seadowg:
OK I've added in a basic nav bar alongside collapsing top bars and proper content padding that works with the floating toolbar:

Screenshot_20260402_154337

Code-wise there's more I'd want to improve, but this is good enough to play with.

Playing with it on device, I do quite like it, although it does feel like the floating bar doesn't require nearly as much padding on any of the sides. We could also try it just sitting on top of the nav bar at full width (with no rounding)?

EDIT: I've just noticed the full screen player layout is super broken. I can fix that, but should affect what we're discussing here really!

formatBCE:
I agree, padding could be reduced. Overall it looks more polished. :)

seadowg:
@formatBCE I've halved the padding. You want me to continue with the floating bar then or try some alternatives?

formatBCE:
> @formatBCE I've halved the padding. You want me to continue with the floating bar then or try some alternatives?

I'd halve it once more, and use floating bar. :)
It looks pretty good!



Is there a way to make bottom bar shorter too? It is HUGE - maybe designs are like this, but it's total waste of space.. :(

seadowg:
> I'd halve it once more, and use floating bar. :)

Sounds good!

> Is there a way to make bottom bar shorter too? It is HUGE - maybe designs are like this, but it's total waste of space.. :(

Yeah, it does seem bigger than it needs to be to me. Easy enough to use a custom component if not.

formatBCE:
I noticed yesterday that toolbar is hiding part of player controls in full view. Did you see that?

P.S. Installed latest one - it looks AWESOME.

Ready to merge @seadowg ? We can add menu items later.

seadowg:
> I noticed yesterday that toolbar is hiding part of player controls in full view. Did you see that?

That should be fixed by https://github.com/music-assistant/mobile-app/pull/194/changes/32f4a8a7ff176df788544266bcb2d4208c59357a. Are you still seeing the problem!

> P.S. Installed latest one - it looks AWESOME.

Aw ace! Yeah, I'm really liking it.

> Ready to merge @seadowg ? We can add menu items later.

There are a few more tweaks to surfaces etc I should make for predictive back - I'll mark the PR as ready to review once I've done that. Agreed that the search/library nav items should be a follow-up PR though. As long as we're not seeing any problems we're probably good to merge this afternoon.

seadowg:
> I'll mark the PR as ready to review once I've done that. Agreed that the search/library nav items should be a follow-up PR though

@formatBCE Ah, sorry I forgot to make it a draft in the first place! It should be good to go now.

formatBCE:
I hope "good to go" means "merge it", because I did. :))

PR #195, opened 2026-04-01
Moved grouping out of the player selection

formatBCE:


formatBCE:
@seadowg  sorry if I'm stepping on your toes. Want to try this solution to group management - it kills me that we have it so far in the rabbit hole.

seadowg:
@formatBCE is your goal here to have player selection **and** grouping be "single-click" actions from everywhere in the app? What about transfer? I definitely like the idea that this app makes player control as easy as possible.

formatBCE:
Yes, as far as I'm concerned judging from amount of requests - people use grouping pretty ofter. So I wanted it to be accessible, and also visible in a glance. Also I'm pissed with empty space there in players button row. :)

Transfer (manual anyways) isn't that often thing. Correct me if I'm wrong - following music, if used, should be automated. It's still 3 clicks away in the app (open player, open queue, click "transfer", same as in web (open queue, click 3 dots [god damn these three dots everywhere, I always am getting confused]< click "transfer").

You still have plans for Transfer?

seadowg:
> Transfer (manual anyways) isn't that often thing. Correct me if I'm wrong - following music, if used, should be automated. It's still 3 clicks away in the app (open player, open queue, click "transfer", same as in web (open queue, click 3 dots [god damn these three dots everywhere, I always am getting confused]< click "transfer").

Right or wrong, this is at least a big assumption that people are using Home Assistant pretty heavily as along-side Music Assistant. Personally I don't, and so I'd prefer to have transfer as a single click. I probably am an anomaly at the moment, but as Music Assistant continues to get better and better, I'd be surprised if we don't see more peeps move over from Roon (or Plex) like myself.

I think the discussion around the question "what should this app's primary actions be?" is fascinating. Most mobile music apps are music browsing focused, and I think there's genuine scope for new ideas here. Going back to Roon, it's probably the closest example in terms of an app that tries to do both player control and browsing well, but I don't think they ever really nailed it on mobile. I'll create an issue we can explore it as a whole there and maybe throw around some dodgy mock-ups and sketches 😆.

PR #197, opened 2026-04-05
Add navigation rail for expanded screens

seadowg:
This replaces the navigation bar with a rail on expanded screens:

Screenshot_20260405_152033

There's a lot more that could be done to optimize/redesign the core layout for expanded screens (fixing the floating bar or embedding it in the rail for example), but this should definitely offer an improvement over a navigation bar on a wider screen.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationScaffold.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationScaffold.kt
@@ -1,0 +1,109 @@
+package io.music_assistant.client.ui.compose.nav
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationScaffold.kt
```diff
@@ -0,0 +1,109 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.window.core.layout.WindowSizeClass
+import io.music_assistant.client.utils.WindowClass
+
+/**
+ * Shows a [NavigationBar] based on [navigationItems] on smaller screens and a [NavigationRail]
+ * instead on expanded and larger.
+ */
+@Composable
+fun AdaptiveNavigationScaffold(
```
seadowg, line 30: Material currently provides a component for this (`NavigationSuiteScaffold`), but it's pretty opinionated, and I found it annoying to work with (it required labels on nav items for example). I think a dead simple custom implementation of this component should be fine for now, and we can always swap out its internals as Jetpack's adaptive navigation components get better.

PR #208, opened 2026-04-08
Add release mode build variant

seadowg:
Compose can perform significantly worse in debug builds. This makes the `selfSignedRelease` build variant non-debuggable and introduces a new `selfSignedDebug` variant that is debuggable. Both use the local debug keystore that Android dev tooling manages.

### DIFF
--- a/.github/workflows/debug-apk-release.yml
+++ b/.github/workflows/debug-apk-release.yml
@@ -44,7 +44,7 @@ jobs:
           EOF
 
       - name: Build debug APK
-        run: ./gradlew :androidApp:assembleSelfSignedRelease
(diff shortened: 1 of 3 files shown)


.github/workflows/debug-apk-release.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           EOF
 
       - name: Build debug APK
-        run: ./gradlew :androidApp:assembleSelfSignedRelease
+        run: ./gradlew :androidApp:assembleSelfSignedDebug
```
seadowg, line 47: We want to keep the APK that CI builds debuggable to make sure it's as easy as possible to grab logs.

formatBCE (approved):

PR #213, opened 2026-04-09
Fix gradient pop-in and contrast

seadowg:
This reworks the way the player pager is rendered so that it's recomposed instead of completely re-rendered when it's expanded and collapsed. This stops the color gradient animations rerunning and generally makes the transition feel a lot smoother.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
```diff
@@ -0,0 +1,401 @@
...
+                                listOf(
+                                    MaterialTheme.colorScheme.surfaceContainerHigh,
+                                    MaterialTheme.colorScheme.surfaceContainerLow
+                                )
+                            )
+                        } else {
+                            Brush.verticalGradient(
+                                listOf(
+                                    MaterialTheme.colorScheme.surfaceContainerHigh,
+                                    dominantColor.copy(alpha = 0.2f)
```
seadowg, line 144: This makes the gradient far less strong, but otherwise we run into contrast problems with dark album covers in light mode.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
```diff
@@ -1,33 +1,82 @@
...
+            .align(Alignment.BottomCenter)
+            .padding(padding)
+            .clip(RoundedCornerShape(clip))
             .fillMaxWidth()
-            .padding(FloatingBarDefaults.padding)
-            .clip(RoundedCornerShape(16.dp))
-            .background(MaterialTheme.colorScheme.surfaceContainerHigh)
-            .clickable { onClick() }
+            .let {
+                if (expanded) {
```
seadowg, line 56: The `FloatingBar` can now expand and collapse which allows us to host content within that that will grow with it as opposed to having to have two different hierarchies for the expanded and collapsed player.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
```diff
@@ -1,33 +1,82 @@
...
+    expanded: Boolean = false,
+    onExpand: (Boolean) -> Unit = {},
+    content: @Composable (expanded: Boolean) -> Unit
+) {
+    BackHandler(enabled = expanded) {
+        onExpand(false)
+    }
+
+    val clip by animateDpAsState(if (expanded) 0.dp else 16.dp)
+    val padding by animateDpAsState(if (expanded) 0.dp else FloatingBarDefaults.padding)
```
seadowg, line 47: Instead of sliding in, this just animates the padding and the clip. This was a whole lot simpler than getting a height change to look good, and I actually really like the "pop" affect - it feels like it grows and then snaps to the side of the window.

formatBCE:
Is this also the reason why queue is collapsing when we switch player? It gets completely recomposed?

seadowg:
@formatBCE good spot! We should be able to fix this now. It looks like the state is just at the wrong level.

seadowg:
@formatBCE done!

formatBCE:
Great!

Issue #199, opened 2026-04-05
Add search to navigation bar


seadowg:
It should be possible to jump to "global" (all providers) search from anywhere in the app showing the navigation bar. 

Search should have its own backstack. Here's some example navigation scenarios:

## Search and playing for an item

1. Launch app
2. Click "Search" in nav bar
3. Type query
4. Click on item
5. Click play on item

The "Search" navigation bar item should be highlighted throughout this flow.

## Backing out of search

1. Launch app
2. Click "Search" in nav bar
3. Type query
4. Click on item
5. Press back

The user should end up on the search page with their query still there. The "Search" navigation bar item should be highlighted throughout this flow.

## Returning home from search

1. Launch app
2. Click "Search" in nav bar
3. Press back

The user should end up on the home screen with the "Home" navigation item highlighted.

## Searching while browsing

1. Launch app
2. Click on item
3. Click "Search" in nav bar
4. Type query
5. Click on item
6. Press back 3 times or click "Home" in nav bar

The user should end up on the item they were on in step 2 with the "Home" navigation item highlighted.

## Returning to search

1. Launch app
2. Click "Search" in nav bar
3. Type query
4. Click on item
5. Click "Home" in nav bar
6. Click "Search" in nav bar

The user ends up on the search screen (the same as after step 2)

seadowg:
@formatBCE as soon as I started playing around with adding a search item I realized there's a bunch of backstack complexity there. Let me know if you disagree with any of the above flows or if you think there are others that we should define before starting work on adding the search item.

formatBCE:
I don't feel like search is that important to have its own stack. Maybe I'm using it wrong, but for me it's just the way to quickly find something and add to library or quickly play and forget. If I play something more or less regularly - I use library.

But modal search won't be feasible with current players placement, because we can't place it on top of everything (user should be able to choose player right before hitting "play" on the item in search). So don't bother much, maybe?

For sure the best thing would be to have Nav host for every item.

seadowg:
> I don't feel like search is that important to have its own stack. Maybe I'm using it wrong, but for me it's just the way to quickly find something and add to library or quickly play and forget. If I play something more or less regularly - I use library.

> But modal search won't be feasible with current players placement, because we can't place it on top of everything (user should be able to choose player right before hitting "play" on the item in search). So don't bother much, maybe?

[...]

Issue #198, opened 2026-04-05
It's hard to perform navigation between related items


seadowg:
When navigating the library/providers, it'd be useful to be able to jump between related items. Here are some actions that I think should be possible:

- [x] Jump from album to album artist(s)
- [ ] Jump from expanded player item to track's album
- [ ] Jump form expanded player item to track's artist

formatBCE:
Looks like item for 3-dot menu

seadowg:
> Looks like item for 3-dot menu

I was thinking you could just click on the text of the artist or album name? That allows multiple artists support as well. I'd also love to just be able to click the album art for the currently playing item and have it just me to the album.

formatBCE:
But how will you articulate this functionality to the user? Menu is explanatory, but with click on element users will have to find this somehow (if ever).

seadowg:
> But how will you articulate this functionality to the user? Menu is explanatory, but with click on element users will have to find this somehow (if ever).

I'd imagine you'd use the [Material Text Button](https://m3.material.io/components/buttons/overview) to signal that it's clickable. The artist name on album pages is consistently clickable in other music apps as well (looking at Spotify, Tidal and Symfonium on mobile and Music Assistant's web UI), so I'd argue it's likely to be an expectation for many users as well. 

Interestingly though, all the apps I tested also have a "Go to artist" in the three dots menu on album pages, so maybe the conclusion is that it's useful to just have both routes for the lowest friction.

formatBCE:
If making artist name clickable, have to make that area to the standard too... Additional hassle.

seadowg:
> If making artist name clickable, have to make that area to the standard too... Additional hassle.

I don't understand what you mean here.

formatBCE:
Just a tap zone size.

seadowg:
> Just a tap zone size.

I think it's already about the right size if not bigger from the extra padding around it. Either way, we can start with the options menu and maybe add the text button links if it's simple or just later on.

We don't have an options/overflow menu on the expanded player however, and I'm not sure adding one makes sense. Do you disagree? Tidal has an interesting alternative to using text links in its "now playing" screen where if you click the track title or artist you get a bottom sheet with options:

<img width="279" height="621" alt="Image" src="https://github.com/user-attachments/assets/9fe7c250-b7ed-48f0-996c-8f94a74c6e0a" />

formatBCE:
Yeah maybe this is the option.
[...]

PR #220, opened 2026-04-12
Backfill smoke and search app tests

seadowg:
Before messing around with navigation for #199 or #198, I wanted to bootstrap "app level" tests so that new ones can be written to drive out the new navigation behaviours. 

These tests start the whole of the shared UI (`App`) and the main "boundary" sits at `ServiceClient` which I've pulled an interface out for so that I could write a stateful fake. The tests use a simple page object pattern to keep the `ComposeTestRule` matcher noise away from the detail of the actual flow being tested, and to make it easier to write new and update tests quickly.

I was hoping to also include the changes needed for #199 in this PR, but I ended up not having enough time, and I'm going to be away most of the next week - probably best to just merge this rather than leaving it hanging!


formatBCE (approved):

formatBCE:
I love that you take your time adding things like this. Good job!

PR #223, opened 2026-04-13
Add search to nav bar

### LINKED ISSUE #199: Add search to navigation bar

seadowg:
Closes #199


seadowg:
@formatBCE I ended up going with a multiple back stack solution here as it surprisingly ended up simpler to implement than the alternative we discussed. It all works well except for the fact that when you do the back gesture (this only affects Android obviously) on search, it closes the app. I think it should return to "Home" (as we discussed in the issue). My guess is that you also won't like this, but I wanted to check before making changes.

EDIT: also it looks like the build might have had a network error fetching a Gradle plugin? Everything seems fine locally.

formatBCE:
> when you do the back gesture (this only affects Android obviously) on search, it closes the app. I think it should return to "Home" (as we discussed in the issue). 
> 

I guess it might be fixed with custom back handler, since we have full access to back stack in Nav3... :)

formatBCE:
> EDIT: also it looks like the build might have had a network error fetching a Gradle plugin? Everything seems fine locally.

Huh looks like Github can't cache AGP 9.1.0?...  IDK why...

seadowg:
> I guess it might be fixed with custom back handler, since we have full access to back stack in Nav3... :)

You'd think so! Annoyingly `NavDisplay` breaks the control we have by always closing the Activity when its backstack is at the last item. There's an [issue filed about this](https://issuetracker.google.com/issues/435410221), but Google doesn't want to fix it because allowing a custom back action on 1 item backstacks messes with predictive back animations for leaving the app.

Google does fortunately have an example of how to implement multiple backstacks with an "exit through home" in their Navigation 3 recipes: https://github.com/android/nav3-recipes/tree/main/app/src/main/java/com/example/nav3recipes/multiplestacks. I can basically just copy their components to get things working as far as I can see. I initially thought that this seemed overly complex (it uses a state wrapper that flattens multiple backstacks into one), but it makes sense given `NavDisplay`'s behaviour. 

Maybe Navigation 4 will fix this 😉.

formatBCE:
Haha wait for Nav4.... :)))

Well let's leave it as is so far.
We always can do that fix later, if very annoying.

formatBCE:
@seadowg  fixed the build.

seadowg:
> @seadowg fixed the build.

Thanks!

> Well let's leave it as is so far.
> We always can do that fix later, if very annoying.

I'll just give the Navigation 3 recipe example a quick try - if it doesn't work "out of the box" I'll abandon it, and we just stick with it as-is.

seadowg:
@formatBCE I went ahead and implemented something fairly similar to the Nav 3 recipe which seems to work pretty well! It doesn't actually end up being massively complex - it just requires using a different setup approach with `NavDisplay` (`entries` of `backstack`). I wouldn't be surprised if we can replace `MultiBackStack` with a framework component in the future. I'll QA it today while I'm working on other things and listening to music in the background.

I think I've stupidly nuked your fixes for the build? I thought you'd changed some config, but I'm guessing you made commits that I've wiped with an `--amend`/`push -f`. Let me know, and I can reincorporate them, or you can add it to `dev` and I can rebase. Sorry about that!

formatBCE:
@seadowg no problem, and done, it's in dev :)

seadowg:
@formatBCE ace! This is ready for review now.

formatBCE:
> @formatBCE ace! This is ready for review now.

Niiiice! Merged it!

PR #234, opened 2026-04-19
Enable TestFlight distribution

teancom:
Make the iOS app uploadable to App Store Connect and usable via TestFlight. A mix of Info.plist keys, build config bumps, Xcode recommended-settings, and targeted cleanup in adjacent files.

Info.plist:
  - CFBundleDisplayName = "Music Assistant" (prettier home-screen label than the smooshed PRODUCT_NAME fallback)
  - ITSAppUsesNonExemptEncryption = false (skips per-upload dialog; app uses only standard HTTPS, which is exempt)
  - NSLocalNetworkUsageDescription: the default server is the .local hostname 'homeassistant.local' (see Defaults.URI); without this key, mDNS resolution silently fails on iOS 14+
  - NSCameraUsageDescription and NSMicrophoneUsageDescription: Apple's static binary scan sees these APIs referenced by WebRTC.xcframework even though the app itself never invokes them
  - Keys alphabetized
  - CFBundleVersion bumped to 2 (build 1 was consumed by a rejected upload that surfaced the missing purpose strings)

gradle.properties:
  - JVM heap 2GB -> 8GB. Release-mode Kotlin/Native compile for iosArm64 OOMs at 2GB (fine for debug Simulator, not for archive).
  - Enable parallel execution and local build caching.

Xcode project:
  - alwaysOutOfDate = 1 on "Compile Kotlin Framework" script phase (silences the "runs every build" warning; Gradle already does up-to-date checking, so declaring outputs would create stale-build hazards).
  - Accept Xcode's recommended settings: DEVELOPMENT_TEAM inherited from project level (removes per-target duplication), String Catalog symbol generation enabled (no-op today, future-friendly), upgrade check bumped.

CarPlay cleanup (while touching adjacent files):
  - CarPlayContentManager: strip 40+ lines of dead scaffolding from fetchRecommendations (outer apiClient.sendRequest whose result was ignored in favor of a direct KmpHelper call), along with the WIP comments that shipped with it.
  - CarPlaySceneDelegate: collapse 7 nearly-identical pushXTemplate() methods into one pushCategoryTemplate(title:fetcher:), driven by a local category table that pairs icons with fetchers directly (replaces the integer-index switch).
  - Remove redundant as? downcast on folder.items (already typed).

Audio codec defaults split per-platform:
  - Codecs is now expect/actual. Android keeps its current behavior (default OPUS, picker shows OPUS/FLAC/PCM); its Concentus-backed OpusDecoder is real and working.
  - iOS defaults to FLAC and hides OPUS from the picker. OpusDecoder.ios is a no-op pass-through stub that returns raw Opus bytes unchanged, so selecting Opus on iOS produces broken audio. A follow-up will implement a real iOS Opus decoder and restore parity.

Verified end-to-end: uploaded 1.0 (2) to App Store Connect, processed to "Ready to Test".

### DIFF
--- a/composeApp/src/androidMain/kotlin/io/music_assistant/client/player/sendspin/audio/Codec.android.kt
+++ b/composeApp/src/androidMain/kotlin/io/music_assistant/client/player/sendspin/audio/Codec.android.kt
@@ -1,0 +1,6 @@
+package io.music_assistant.client.player.sendspin.audio
+
+actual object Codecs {
+    actual val default: Codec = Codec.OPUS
+    actual val list: List<Codec> = listOf(Codec.OPUS, Codec.FLAC, Codec.PCM)
+}
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/Codec.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/Codec.kt
@@ -17,9 +17,9 @@ enum class Codec(
     }
 }
 
-object Codecs {
-    val default = Codec.OPUS
-    val list: List<Codec> = listOf(Codec.OPUS, Codec.FLAC, Codec.PCM)
+expect object Codecs {
+    val default: Codec
+    val list: List<Codec>
 }
 
 fun codecByName(name: String): Codec? =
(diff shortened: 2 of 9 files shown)


formatBCE (approved):
Great changes! Thank you!

PR #236, opened 2026-04-20
feat(ios): implement OAuth login

teancom:
Wire up OAuth support on iOS so users with HA-linked Music Assistant servers can authenticate. The iOS OAuthHandler implementation already existed but was never instantiated, leaving OAuth-based providers inoperable with "OAuth not supported on this platform."

- KmpHelper exposes authManager to Swift.
- Info.plist registers `musicassistant://`, the URL scheme that AuthenticationViewModel hardcodes as the OAuth return URL.
- iOSApp.swift assigns OAuthHandler() in ContentView.onAppear (after Koin init) and parses the callback via .onOpenURL, forwarding the code to authManager.handleOAuthCallback.
- Cold-launch deep links (URL arrives before Koin ready) are buffered and replayed by the same onAppear.

Mirrors Android's MainActivity.onCreate wire-up.


formatBCE (approved):
Thanks for adding OAuth!

PR #237, opened 2026-04-20
chore(ios): hash-gate gradle invocation in Compile Kotlin Framework

teancom:
Skip the ./gradlew call in the Compile Kotlin Framework script phase when Kotlin sources haven't changed since the last successful run, avoiding the ~30s cold-daemon startup Gradle pays on every Xcode invocation. Saves roughly 30s per no-op re-archive (e.g. bump CFBundleVersion and re-archive). Has no effect when sources did change — gradle runs normally, gate adds <100ms.

Timings measured via MCP xcodebuild on this branch (Debug, iphoneos):

  Cold build (empty cache):          207.9s
  No-op build (sources unchanged):     2.9s
  Real source change (warm daemon):   33.3s

Hash covers composeApp sources, compose resources, and key build- config files. Stored under DERIVED_FILE_DIR so it's scoped per- configuration and per-SDK (no cross-contamination between Debug/ Release, device/simulator). alwaysOutOfDate=1 stays — Xcode still runs the script every build; the gate is inside the script.

Also drop the kotlin.native.cacheKind=none override. It was added as a Crashlytics workaround; Crashlytics isn't integrated.


teancom:
@formatBCE I wanted to make sure you noticed that I removed `kotlin.native.cacheKind=none` in particular. From what I can tell it was Crashlytics related, and things seem to work fine (better) with it gone. But if I'm missing something / context / whatever, please let me know.

formatBCE:
> @formatBCE I wanted to make sure you noticed that I removed `kotlin.native.cacheKind=none` in particular. From what I can tell it was Crashlytics related, and things seem to work fine (better) with it gone. But if I'm missing something / context / whatever, please let me know.

Nice.
it's not needed so far - I'm still thinking on which platform to use for crashes/logs. Probably it's up to OHF to decide.

formatBCE (approved):

PR #244, opened 2026-04-22
Add library to nav bar

seadowg:
This adds a Library nav bar item (with its own back stack like Search) and removes the library row from Home. I've also made it so tapping on a nav bar item when it's "active" will reset it so you can easily return to Home, Library or Search when you're deep in a stack.


androidApp/src/test/resources/robolectric.properties
```diff
@@ -1,2 +1,2 @@
 sdk=34
-application=io.music_assistant.client.support.TestApplication
```
seadowg, line 2: I wanted to make the test setup more explicit and less reliant on the Robolectric tricks (knowledge of which is becoming rarer and rarer). That will also hopefully make them easier to move to multiplatform tests once the tooling for that is more mature.

formatBCE: I really glad we have you here! I love tests coverage, but hate to do this important job... Shame on me!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -137,6 +140,7 @@ fun MainNavigationRoot(
 
     val backStacks = listOf(
         rememberMainNavBackStack(MainNav.Landing),
+        rememberMainNavBackStack(MainNav.Library(MediaType.ARTIST)),
```
formatBCE, line 143: We need to make this setting adjustable. Many guys use albums as their main navigation anchor. Now they need two clicks.

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
seadowg: Very much agree! I hadn't really though about that until writing this line. I think also being able to remove ones you don't use (even if they're not empty) would be nice as well.

androidApp/src/test/resources/robolectric.properties
seadowg: Tests are hard. Glad you're liking it though! I feel like actually having tests in Android/iOS app codebases is pretty rare looking at the industry as a whole, so sometimes people have a frosty reaction to them. I was indoctrinated into TDD early on in my career and now find it genuinely uncomfortable writing code without it which I think gives me that extra motivation to get this annoying bootstrapping done 😆. 

The more fleshed out the page objects and `FakeServiceClient` get, the easier it'll be to write new tests hopefully!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
formatBCE: Some setting with reorder able list with switches?

formatBCE: Also I had bug from scratch there: if, for example, user goes to Albums, open Album, then clicks Back - he lands on Artists. Args library item should be replaced with current chosen item.

seadowg: > Some setting with reorder able list with switches?

Yeah, I can imagine that! I think we'll probably want an "Interface" settings page with sub-pages for customizing library, item pages, home etc. in future. Maybe the library one is a good starting point though? Might just be worth adding an issue now, but it feels like it's not top priority yet.

PR #249, opened 2026-04-23
Add "Go to artist" option for albums

seadowg:
Work towards #198

As discussed in the issue, this adds navigation from an album to artists via the overflow ("More") menu.


androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
```diff
@@ -198,23 +225,28 @@ class FakeServiceClient(private val settingsRepository: SettingsRepository) : Se
         TODO("Not yet implemented")
     }
 
-    fun addToLibrary(vararg albums: AppMediaItem.Album) {
-        this.albums.addAll(albums)
+    fun addToLibrary(vararg items: ServerMediaItem) {
```
seadowg, line 228: I realized it was kind of weird/awkward to be talking to `FakeServiceClient` with `AppMediaItem` instances as it didn't deal in them at all.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemHeader.kt
```diff
@@ -204,10 +208,20 @@ private fun ItemOverflow(
                     onClick = onToggleViewMode
                 )
             )
+
+            if (goToArtist != null) {
+                add(
+                    OverflowMenuOption(
+                        title = "Go to artist",
+                        icon = Icons.Default.Person,
```
seadowg, line 216: I couldn't access `Icons.Default.Artist` for some reason and I didn't want to go down the rabbit hole of the multiplatform lib vs the androidx one right now!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -332,6 +332,11 @@ private fun ItemContent(
                 onToggleViewMode = onToggleViewMode,
                 libraryActions = libraryActions,
                 playlistActions = playlistActions.takeIf { item !is AppMediaItem.Genre },
+                goToArtist = if (item is AppMediaItem.Album && item.artists.isNotEmpty()) {
+                    { onNavigateClick(item.artists[0]) }
```
seadowg, line 336: This UX needs to get a bit more complicated for albums with more than one artist, but I couldn't personally find any examples in my own collection (I think Navidrome supports it though) and I don't think most streaming services even support the concept. I think this is something we'd want to add later.

Generally for me right now, splits are all under a single "album artist" and tracks have the specific artists. Compilations tend to use a placeholder album artist like "Various artists". It does look to me like Music Assistant web doesn't currently like that setup as it doesn't have an "appears on" for artists - the tracks in splits/compilations appear in "Tracks", but the album is only present for the album artist.

androidApp/src/test/kotlin/io/music_assistant/client/support/pages/HomePage.kt
```diff
@@ -3,14 +3,16 @@ package io.music_assistant.client.support.pages
...
 class HomePage(composeTestRule: ComposeTestRule) : ComposePage(composeTestRule) {
     override fun assert() {
         composeTestRule.onNodeWithText("Home").assertIsDisplayed()
         assertNavBar(items = listOf("Home", "Library", "Search", "Settings"), selected = "Home")
     }
 
-    fun clickOnMedia(name: String): MedaItemPage {
-        return clickOnMedia(name, "Home")
+    fun clickOnMedia(item: ServerMediaItem): MediaItemPage {
+        return clickOnMedia(item, "Home")
```
formatBCE, line 15: It duplicates the line for me... ?

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/AppMediaItemFixtures.kt
```diff
@@ -24,7 +24,11 @@ object AppMediaItemFixtures {
             image = null,
             version = version,
             year = null,
-            artists = listOf(artist)
+            artists = if (artist != null) {
```
formatBCE, line 27: I tend to use Kotlin `artist?.let { list(it) } ?: empty list()`

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -273,7 +273,7 @@ private fun ItemChildren(
                     onToggleViewMode = onToggleViewMode,
                     onAlbumsSortChanged = onAlbumsSortChanged,
                     onPlayableItemsSortChanged = onPlayableItemsSortChanged,
-                    contentPadding = contentPadding,
```
formatBCE, line 276: Oh no...

I love trailing commas. They let future additions look much prettier in git diff. :)

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
seadowg: Oh sorry! I don't have a vendetta against them. I had added and then removed a parameter while working on this and probably removed this by accident.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/AppMediaItemFixtures.kt
seadowg: I sometimes find the "elvis" operator can add a lot of noise, so I do often end up using explicit `if-else` instead, but you're probably right that it would have been cleaner here.

androidApp/src/test/kotlin/io/music_assistant/client/support/pages/HomePage.kt
seadowg: Huh yeah that's a mistake. I probably hit CMD + D by accident. I've actually never noticed that Kotlin's compiler doesn't have an "unreachable code" error - it must just silently remove it.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/AppMediaItemFixtures.kt
formatBCE: Yeah, nested Elvis operators are pretty messy.

PR #262, opened 2026-04-24
fix(api): harden decode boundaries and RpcEngine concurrency

teancom:
I was going through Test Flight stuff and noticed that we've been collecting crash reports and feedback from our initial wave of users, which is *lovely*. Here is a collection of fixes for the four types of crashes that we've seen so far.

ServerPlayer and ServerQueue declared required fields that the server sometimes omits, and their enum fields (PlayerType, PlayerState, RepeatMode) threw SerializationException on unknown wire variants. Non-identity fields now carry safe defaults, and myJson opts into coerceInputValues so unknown enum variants degrade to the default (or null) instead of throwing. Answer.resultAs, Event.event, and the ServerInfo handshake catch SerializationException at the decode boundary so schema drift stays local to the affected RPC. Both platforms benefit: on Kotlin/Native the uncaught exception aborted the whole process, on JVM it would have killed the background coroutine and dropped events.

RpcEngine.pendingResponses and partialResults were plain HashMaps mutated from both the request-issuing coroutines and the websocket message-collector coroutine. HashMap isn't concurrency-safe on either platform; Kotlin/Native surfaced it as
ArrayIndexOutOfBoundsException out of HashMap#addKey, but the JVM was latently vulnerable to the same bucket-array corruption. Both maps are now AtomicReference<Map> with copy-on-write CAS updates, matching the pattern already used elsewhere in the codebase.

Tests (commonTest, 35 new cases across five files) cover minimal ServerPlayer/ServerQueue payloads, unknown enum coercion, list decode surviving sparse entries, Answer/Event behaviour under decode failure, plus two stress tests that hammer RpcEngine on Dispatchers.Default to exercise the AtomicReference CAS loops.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/Answer.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/Answer.kt
@@ -1,6 +1,8 @@
 package io.music_assistant.client.api
 
+import co.touchlab.kermit.Logger
 import io.music_assistant.client.utils.myJson
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.decodeFromJsonElement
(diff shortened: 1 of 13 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/api/Answer.kt
```diff
@@ -11,5 +13,26 @@ data class Answer(
...
+     * [SerializationException] on a background coroutine dispatcher aborts
+     * the whole Kotlin/Native process, so schema drift must stay local to the
+     * affected RPC rather than crash the app.
+     */
+    inline fun <reified T : Any> resultAs(): T? {
+        val payload = result ?: return null
+        return try {
+            myJson.decodeFromJsonElement<T>(payload)
+        } catch (e: SerializationException) {
+            Logger.withTag("Answer").w(e) {
```
formatBCE, line 31: Would be great to have field for this logger to avoid calling "wothTag" on every catch.

formatBCE (approved):

PR #268, opened 2026-04-25
Don't reanimate colors when switching player pagers

seadowg:
I made this change while working on #198. I didn't get as much time as I thought, so wanted to get this in instead of it waiting for me to finish the other bits.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
@@ -48,9 +48,6 @@ import androidx.compose.ui.unit.DpSize
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import io.music_assistant.client.data.model.client.AppMediaItem
-import musicassistantclient.composeapp.generated.resources.Res
-import musicassistantclient.composeapp.generated.resources.*
-import org.jetbrains.compose.resources.stringResource
 import io.music_assistant.client.data.model.client.PlayerData
 import io.music_assistant.client.player.sendspin.SendspinState
 import io.music_assistant.client.ui.compose.common.action.PlayerAction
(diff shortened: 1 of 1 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
```diff
@@ -111,6 +111,14 @@ internal fun PlayersPager(
         )
     }
 
+    val dominantColors = playerDataList.associateWith {
```
seadowg, line 114: This also means most colors will likely be fetched before the user sees them resulting in less visible pop-in.

formatBCE, line 114: Probably would be great to have some [URL, color] map cached somewhere? Now with player data as key it will be still recalculated every time player data hash is changed (which is pretty much every time). :) Or do I miss something?

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
seadowg: So on `dev` you get a `DominantColorState#updateFrom` every single time you flick between players with artwork. These changes reduce that to just a call for every player with artwork when the app opens and then when new artwork is presented so it does seem more efficient right now - I think it's worth merging as-is.

That said, I think it's really hard to follow what's happening when you get to this level of `remember`/`State`/`LaunchEffect` combinations! I do think having a component that handles this away from Compose with (maybe a ViewModel hosting an LRU cache for example) would be much nicer. Maybe we want an `ArtworkViewModel` that handles downloading (and caching) artwork as well as exposing extracted colors? From an actual product perspective, I think the story would be "Artwork shoudn't be refetched when I play an album I just played".

formatBCE: VM is good thought.

seadowg: > VM is good thought.

Ace. I can write an issue up.

PR #277, opened 2026-04-25
fix(ios/carplay): black screen on launch

teancom:
A cold launch from CarPlay connects only the CPTemplateApplicationScene;
SwiftUI's WindowGroup scene never connects, so ContentView.onAppear
never fires. Koin was initialized inside MainViewController's configure
block (driven by ContentView), so on this path Koin never started and
CarPlaySceneDelegate waited forever for KMPReadyNotification — no
setRootTemplate call, head unit shows black.

Move bootstrap to iOSApp.init(), which runs on the main thread before
any scene-delegate didConnect. Extract bootstrapKmp() in
MainViewController.kt as the single entry point, backed by a
Unit-by-lazy holder so callers from both Swift (iOSApp.init) and the
ComposeUIViewController configure block are safe — startKoin would
otherwise throw on the second call.

Drop the now-dead KMPReadyNotification observer branch from
CarPlaySceneDelegate. KmpState/KMPReadyNotification remain in
iOSApp.swift for SiriIntentHandler, which can be invoked outside the
scene lifecycle.

Update .claude/carplay.md to match the new architecture.


teancom:
Oh, I also added a build version bump so I can release this with Test Flight once merged. Fixing CarPlay support felt like a big enough change to warrant it.

formatBCE (approved):

PR #283, opened 2026-04-27
Add refresh for home screen

seadowg:
There's no way to refresh the Home screen once the app is loaded. This is annoying if you add an album and then want to play it.


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/AppMediaItem.kt
```diff
@@ -165,6 +165,16 @@ abstract class AppMediaItem(
             "favorite_playlists" -> MediaType.PLAYLIST
             else -> null
         }
+
+        override fun equals(other: Any?): Boolean {
```
seadowg, line 169: The `equals` implementation in `AppMediaItem` was preventing recomposition when the `RecommendationsFolder` was updated as it didn't include `items`.

I think we should really convert all the `AppMediaItem` subclasses to `data` classes which requires a little rework of the structure - better as a follow-up I can come back with.

formatBCE (commented):
Why removing quick search? 
it's so helpful! With dynamic loading it's the only way to get over the extensive library.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
```diff
@@ -126,10 +131,18 @@ fun HomeScreen(
...
 
 @Composable
-private fun LandingPageTopBar(scrollBehavior: TopAppBarScrollBehavior) {
+private fun LandingPageTopBar(scrollBehavior: TopAppBarScrollBehavior, onRefresh: () -> Unit) {
     TopAppBar(
         title = { Text(stringResource(Res.string.nav_home)) },
-        scrollBehavior = scrollBehavior
+        scrollBehavior = scrollBehavior,
+        actions = {
+            IconButton(onClick = onRefresh) {
```
seadowg, line 139: I also considered pull-to-refresh for this, but I think we'd always want a button anyway for accessibility reasons. It'd now be easy to layer that on top!

seadowg:
> Why removing quick search?

Just as in the PR description it seemed like it was no pretty confusing to have search available in two places. I don't think it's particularly clear what "scope" is in play.

> With dynamic loading it's the only way to get over the extensive library.

Can't you use the "In library only" filter?

If you feel strongly about keeping it, I can just remove that commit from the PR!

formatBCE:
> 
> If you feel strongly about keeping it, I can just remove that commit from the PR!

IDK it's really the thing that I use daily.
Using global search is possible, but 1 - slower, 2 - counterintuitive.
If it does search some shit - we need to fix it in MA, not remove functionality :)

seadowg:
> If it does search some shit - we need to fix it in MA, not remove functionality :)

Oh to be clear I meant just I was trying to use it without thinking about it. Cool I'll remove that commit for now.

androidApp/src/test/kotlin/io/music_assistant/client/feature/SmokeTest.kt
```diff
@@ -1,42 +0,0 @@
-package io.music_assistant.client.feature
```
seadowg, line 1: This wasn't really needed any longer as other tests cover the same paths.

formatBCE (approved):

PR #298, opened 2026-04-28
feat(ios): expand Siri integration

teancom:
With this, you can now say "Hey Siri, in Music Assistant play <album> by <artists>" and sometimes it works! (The 'sometimes' seems to be mostly based on whether or not Siri understood you correctly, decided that what you were saying was an album or artist or track or something else. Whenever it *did* understand me, it seemed to work).

You can also hold down the voice assistant button and say 'Play <artist> in Music Assistant' with the same "Works at least some of the time" reliability that Siri is known for.

Also, we now ask for Siri permission on first launch the way we do for 'local network' permission. It's a one-time prompt, if the user says No that's fine, we don't ask again (they can turn it on manually via the Settings -> App -> Music Assistant -> Siri path).

Robot sidekick description to follow:

Adds donation plus three intent handlers (play, search, affinity) and the
Info.plist keys (INSupportedMediaCategories, NSUserActivityTypes) Siri's
routing requires for non-Apple-Music apps to appear as media destinations.
Resolution uses token-overlap similarity scoring against name+subtitle
(with diacritic folding) to override MA's server-side relevance when wrong,
and a no-disambig policy on PlayMedia to avoid Siri's tap-loop failure
mode. Donations are namespaced by server id (newly exposed via
KmpHelper.getServerId from ServerInfo.serverId) so the same item across
MA instances doesn't fragment Siri's prediction model. Affinity disambig
is capped at five, and Search now returns cache-derived items rather than
echoing back whatever Siri substituted from its own catalog.

Also folds in the prerequisites we hit while testing: bootstrapKmp() moved
into iOSApp.init() so a cold "Hey Siri, play X" that connects only the
CarPlay scene still gets Koin up; OSLog subsystems use
Bundle.main.bundleIdentifier so diagnostic logs survive (the previous
underscore-vs-hyphen mismatch was silently dropping every Siri-category
line on real devices); and KmpState.isReady is lock-protected for the
cross-thread read intent handlers do. See .claude/carplay.md, including
the new note distinguishing intent-handled phrases from MPRemoteCommand-
routed transport commands like "next track" and "pause."


formatBCE (approved):

formatBCE:
Jealous... So easy in iOS?

PR #299, opened 2026-04-29
Optimization to dominant color extraction and Coil caching.

### LINKED ISSUE #286: Reduce requests needed for artwork
The app currently makes network requests for artwork for both displaying artwork and for calculating the dynamic color for background/controls. Added to that the state for the dynamic colors is pretty hard to follow. There is current caching for image files and colors handled by `AsyncImage` and `DominantColorState`, but this is all handled under the hood of those frameworks and isn't coordinated.

The could all be optimized (and simplified) by introducing a component that's responsible for fetching artwork and calculating colors (most likely a ViewModel) based on image URLs. It looks like we could just use Coil to implement it to make use of its built-in caching. Something like:

```
fun fetchImage(url: String): StateFlow<ImageState?> {
    val stateFlow = MutableStateFlow<ImageState?>(null)

    val request = ImageRequest.Builder(context)
            .data(imageUrl)
            .target(
                onSuccess = { drawable ->
                    val color = Palette.from(bitmap).generate().getDominantColor()
                    stateFlow.value = ImageState(bitmap, color)
                }
            )
            .build()

    imageLoader.enqueue(request)
    return stateFlow
}
```

formatBCE:
Fixes #286

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColor.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColor.kt
@@ -16,8 +16,22 @@ import androidx.compose.runtime.produceState
 import androidx.compose.runtime.remember
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.luminance
-import coil3.compose.LocalPlatformContext
-import org.koin.compose.koinInject
+
+/**
+ * Theme-independent extraction result kept in [DominantColorViewModel]'s cache.
+ * Both tint variants are pre-computed so consumers select cheaply by surface luminance.
+ */
+data class ExtractedColors(
+    val dominant: Color,
+    val tintOnDark: Color,
+    val tintOnLight: Color,
+)
+
+/**
+ * Suspending fetcher used by [rememberAnimatedPlayerColors] — supplied by the screen
+ * so the composable doesn't depend on Koin and is trivially testable with a fake.
+ */
+typealias ExtractedColorsFetcher = suspend (imageUrl: String) -> ExtractedColors?
 
 /**
  * Dominant color extracted from artwork plus its theme-adjusted control tint.
@@ -33,16 +47,15 @@ data class PlayerColors(
 fun rememberAnimatedPlayerColors(
     imageUrl: String?,
     fallback: Color,
+    fetchColors: ExtractedColorsFetcher,
 ): State<PlayerColors> {
-    val viewModel: DominantColorViewModel = koinInject()
-    val context = LocalPlatformContext.current
     val onDark = MaterialTheme.colorScheme.surface.luminance() < 0.5f
 
-    val extracted by produceState<DominantColorViewModel.ExtractedColors?>(
+    val extracted by produceState<ExtractedColors?>(
         initialValue = null,
         key1 = imageUrl,
     ) {
-        value = imageUrl?.let { viewModel.getColors(context, it) }
+        value = imageUrl?.let { fetchColors(it) }
     }
 
     val targetDominant = extracted?.dominant ?: fallback
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColor.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColor.kt
@@ -9,44 +9,59 @@ import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.produceState
+import androidx.compose.runtime.remember
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.luminance
-import com.kmpalette.extensions.network.rememberNetworkDominantColorState
-import io.ktor.http.Url
+import coil3.compose.LocalPlatformContext
+import org.koin.compose.koinInject
+
+/**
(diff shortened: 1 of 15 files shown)


formatBCE:
@seadowg feel free to optimize/refactor while I'm hitting the hay :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
```diff
@@ -30,21 +30,22 @@ class DominantColorViewModel : ViewModel() {
         val tintOnLight: Color,
     )
 
-    private val mutex = Mutex()
-    private val cache = LinkedHashMap<String, ExtractedColors>()
+    private val cache = LruCache<String, ExtractedColors>(MAX_CACHE_SIZE)
```
seadowg, line 29: This switches the behaviour as well obviously, but I think LRU is probably more appropriate than FIFO here as it increases the changes of colors in the player bar not getting kicked out by browsing.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
```diff
@@ -30,21 +30,22 @@ class DominantColorViewModel : ViewModel() {
         val tintOnLight: Color,
     )
 
-    private val mutex = Mutex()
```
seadowg: `LruCache` is thread safe, so I don't think we need any synchronization in here any longer. There's a "risk" that we make multiple requests for the same key (URL), but that was already the case, and it's not going to result in any bad state.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private suspend fun extract(context: PlatformContext, url: String): ExtractedColors? {
+        val request = ImageRequest.Builder(context)
+            .data(url)
+            .disableHardwareBitmaps()
+            .memoryCachePolicy(CachePolicy.ENABLED)
+            .memoryCacheKey(url)
+            .build()
+        val result = SingletonImageLoader.get(context).execute(request) as? SuccessResult
```
seadowg, line 47: So because we're using `SingletonImageLoader` do we end up using the same memory cache for the actual images as `AsyncImage`?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+class DominantColorViewModel : ViewModel() {
+    data class ExtractedColors(
+        val dominant: Color,
+        val tintOnDark: Color,
+        val tintOnLight: Color,
+    )
+
+    private val cache = LruCache<String, ExtractedColors>(MAX_CACHE_SIZE)
+
+    suspend fun getColors(context: PlatformContext, imageUrl: String): ExtractedColors? {
```
seadowg, line 31: Personally, I'd drop the `suspend` call stack here and have `getColors` return a `StateFlow` instead. That makes `ViewModel` actions easier to test as you can just pass in dispatchers to control asynchronous work and from a personal perspective I think it makes it easier to reason about. I won't make that change unless you're ok with it though!

seadowg, line 31: Can Koin inject `context` (from the `Application`) at the constructor level instead of it needing to be a method param?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColor.kt
```diff
@@ -9,44 +9,59 @@ import androidx.compose.animation.animateColorAsState
...
+fun rememberAnimatedPlayerColors(
     imageUrl: String?,
     fallback: Color,
-): State<Color> {
-    val dominantColorState = rememberNetworkDominantColorState(
-        defaultColor = fallback,
-        defaultOnColor = Color.White,
-    )
+): State<PlayerColors> {
+    val viewModel: DominantColorViewModel = koinInject()
```
seadowg: I'd be tempted to hoist the `viewModel` up to the "screen" level (`MainNavRoot`) and then pass it down as an interface (`ColorFetcher` or something) or a function (`(String) -> ExtractedColors?`). Otherwise, any test of a component that calls this would need a Koin environment set up. What do you think?

formatBCE: It's either that or that. Koin is OK with tests, but it's easier to just pass lambda. :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
formatBCE: There's no PlatformContext in Shared module, so we'd have to move VM to platform-specific modules.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
formatBCE: Yes, it's using same cache

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DominantColorViewModel.kt
formatBCE: I tried it - and it turns pretty gibberish to read...

seadowg: Gotcha!

PR #307, opened 2026-04-30
Fixed width to volume number, stopped slider jumping

### LINKED ISSUE #300: Bug / iOS - After addition of volume percentage for external players in volume bar, volume bar width differs based on number width
LOVE the new visible percentage when setting volume on external player. But, I think the width of the volume bar is now variable/calculated based on the width of the numbers next to it. Shouldn't the bar have a fixed width (percentage of viewwidth based maybe?) instead of calculated? Currently the calculated width feels a bit wonky. (See volume bar in screenrecording below).

https://github.com/user-attachments/assets/3577e6ed-a53d-46f8-8a63-daee3c98f0a5

formatBCE:
Fixes #300


remon1496:
Just found this one by accident.

IMG_5126

formatBCE:
Yup aware of that. Bummer that we need fixed width there because of slider.

PR #310, opened 2026-04-30
feat(player-selection): persist user choice

teancom:
The iOS app would often switch to whatever player came first in the list of available players, vs. using what the user had selected last time, on launch. We now persist the user's selection.

Robot's detailed explanation:

Replace the auto-select coroutine that pushed first-available fallbacks
into `_selectedPlayerId` with a pure resolver over (visible players,
user choice). The effective selection is now a `combine` over
`_playersData` and `_userSelectedPlayerId`, computed by
`resolveSelectedPlayerId`:

  1. user choice if it appears in the visible list — persisted explicit
     pick wins
  2. first visible player — fallback when no user choice or when the
     user's choice is offline
  3. user choice as a last resort — preserves the persisted value while
     the player list is loading or empty so the UI holds steady across
     the gap

Only `selectPlayer` writes `_userSelectedPlayerId`, and a single
collector in `MainDataSource.init` mirrors non-null values into
`SettingsRepository.lastSelectedPlayerId` (backed by
`last_selected_player_id` in platform settings — SharedPreferences on
Android, NSUserDefaults on iOS). Restoration on startup is automatic:
the upstream flow seeds from `settings.lastSelectedPlayerId.value` and
the resolver re-evaluates on every player-list emission, so a player
that joins after launch (e.g. local sendspin finishing registration)
is picked up without state-machine help.

Tests. `SelectedPlayerResolutionTest`: all three resolver
branches plus edge cases (empty list, user choice not in list, list
growing post-launch, explicit user pick takes priority over a
sendspin-local that landed first in the order).

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
@@ -178,7 +178,36 @@ class MainDataSource(
             .map { it.data.any { data -> data.queueInfo?.currentItem != null } }
             .stateIn(this, SharingStarted.Eagerly, false)
 
-    private val _selectedPlayerId = MutableStateFlow<String?>(null)
+    /**
+     * Persisted user choice. Restored from settings on startup and written
+     * only by [selectPlayer]; the persistence collector in [init] mirrors
+     * non-null values into [SettingsRepository.lastSelectedPlayerId] for
+     * the next app launch.
+     */
+    private val _userSelectedPlayerId =
+        MutableStateFlow(settings.lastSelectedPlayerId.value)
+
+    /**
+     * Effective selection consumed by the rest of the data source and UI.
+     * Derived from the current player list and the user choice via
+     * [resolveSelectedPlayerId] — pure function, re-evaluated on every
+     * input change. No state machine pushes a fallback into the upstream
+     * flow; the resolver computes it on the fly.
+     */
+    private val _selectedPlayerId: StateFlow<String?> =
+        combine(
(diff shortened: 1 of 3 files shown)


formatBCE (commented):
This thing is huge, and I have strong feeling that this is because of convoluted CarPlay implementation...
The player exposer to CarPlay should be only `MainDataSource.localPlayer`. That's it, in this case CarPlay can co-exist with main application flawlessly.
If other players are exposed to the CarPlay, we will need to think on what is the correct thing to do...

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -1750,7 +1775,28 @@ class MainDataSource(
...
+         *  3. [userChoice] as a last resort — preserves the persisted value
+         *     while the player list is loading or empty so the UI holds
+         *     steady across the gap.
+         */
+        internal fun resolveSelectedPlayerId(
+            visiblePlayerIds: List<String>,
+            userChoice: String?,
+        ): String? {
+            if (userChoice != null && userChoice in visiblePlayerIds) return userChoice
+            return visiblePlayerIds.firstOrNull() ?: userChoice
```
formatBCE, line 1799: This logic is flawed. 
If userChoice is not null, and visiblePlayerIds are empty - we will return userChoice. I would remove last part of elvis operator and just return null there.
Doesn't hurt, while players are empty - but redundant and wrong anyways. :) We never should return value that isn't in the list.

formatBCE (approved):

PR #312, opened 2026-05-01
fix(queue): drop stale queue event replays; handle queue_added

teancom:
I saw a lot of out-of-order and replayed Queue events when I was on cellular, and using my steering wheel buttons and Siri to control playback. It ended up with a lot of weird UI stuff, showing some track for a split second then snapping over to another one, all while playing the same song. This uses the time sent from the server as a monotonic-ish (and DST-immune) guide to what comes first so we do better at displaying what the server thinks we should be displaying.

Robot explanation:

The MA server occasionally replays QueueUpdated/QueueItemsUpdated events that predate fresher state we already received (observed around Siri "next track" handoffs and rapid queue mutations); the playhead would snap backward or a prior track would briefly return.

`QueueInfo.isStaleReplay` drops events whose `elapsedTimeLastUpdated` predates what we have for the same queue id, gated through a single `gateOrSkip` helper across `QueueUpdated`, `QueueItemsUpdated`, and `QueueAdded`. `_queueInfos` is the single source of truth: optimistic shuffle/repeat toggles bump the stamp to client wall clock and notify `MainDataSource` via `LocalPlayerRepository.onOptimisticQueueChange`, so a stale replay after a toggle can't pass the gate. `QueueTimeUpdatedEvent` carries no server stamp and isn't gated.

QueueAdded was previously dropped at `Event.fromJsonElement`; the new handler upserts into `_queueInfos` and forwards to the local-player repository, fixing an empty Now Playing on a freshly connected player.

Tests: 10 cases in `QueueInfoStalenessTest` + queue_added parse test.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalPlayerRepository.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalPlayerRepository.kt
@@ -14,6 +14,7 @@ import io.music_assistant.client.player.MediaPlayerController
 import io.music_assistant.client.settings.SettingsRepository
 import io.music_assistant.client.ui.compose.common.DataState
 import io.music_assistant.client.ui.compose.common.action.PlayerAction
+import io.music_assistant.client.utils.currentTimeMillis
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.IO
@@ -40,6 +41,18 @@ class LocalPlayerRepository(
 
     private val _localPlayerData = MutableStateFlow<PlayerData?>(null)
     val localPlayerData: StateFlow<PlayerData?> = _localPlayerData.asStateFlow()
+
+    /**
+     * Fired whenever an optimistic queue mutation produces a new [QueueInfo]
+     * for the local player. [MainDataSource] sets this in its `init` block to
+     * mirror the bumped `elapsedTimeLastUpdated` into its own `_queueInfos`
+     * store, which is the single source of truth consulted by the staleness
+     * gate. Without the mirror, a stale server replay arriving after an
+     * optimistic toggle could pass the gate (it'd compare against the old
+     * server stamp, not the bumped optimistic one) and clobber the user's
+     * change.
+     */
+    var onOptimisticQueueChange: ((QueueInfo) -> Unit)? = null
 
     private val commandQueueMutex = Mutex()
     private val commandQueue = mutableListOf<QueuedEntry>()
@@ -145,6 +158,8 @@ class LocalPlayerRepository(
     }
 
     fun onServerQueueUpdate(queueInfo: QueueInfo) {
+        // Staleness gating happens upstream in [MainDataSource.gateOrSkip]
+        // before we're called, so this just absorbs the admitted event.
         _localPlayerData.update { current ->
             current?.copy(
                 queue = DataState.Data(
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalPlayerRepository.kt
```diff
@@ -41,6 +41,18 @@ class LocalPlayerRepository(
...
+     * Fired whenever an optimistic queue mutation produces a new [QueueInfo]
+     * for the local player. [MainDataSource] sets this in its `init` block to
+     * mirror the bumped `elapsedTimeLastUpdated` into its own `_queueInfos`
+     * store, which is the single source of truth consulted by the staleness
+     * gate. Without the mirror, a stale server replay arriving after an
+     * optimistic toggle could pass the gate (it'd compare against the old
+     * server stamp, not the bumped optimistic one) and clobber the user's
+     * change.
+     */
+    var onOptimisticQueueChange: ((QueueInfo) -> Unit)? = null
```
formatBCE: This should be a Channel, and MainDataSource should consume from it. Or at least the StateFlow...

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -1030,6 +1047,30 @@ class MainDataSource(
...
+     *
+     * `_queueInfos` is the single source of truth: server-event handlers
+     * write to it on admit, and [LocalPlayerRepository.onOptimisticQueueChange]
+     * mirrors optimistic-bump stamps into it via the callback wired in [init].
+     * That removes the need for a cross-store merge — the gate only has to
+     * look in one place.
+     */
+    private fun gateOrSkip(data: QueueInfo, label: String): Boolean {
+        val existing = _queueInfos.value.find { it.id == data.id }
+        if (QueueInfo.isStaleReplay(incoming = data, existing = existing)) {
```
formatBCE: Instead of static function I'd make
```
fun QueueInfo.isBefore(other: QueueInfo) {
...
}
```
 And use it here like `data.isBefore(existing)`.
It's basically the same, but semantically much better.

Overall, I'd suggest using system prompt (e.g. persona or skill) with instructions to follow SOLID and KISS. It helps to harness the beast. :)
If you want, I can give my humble prompt. It really makes difference for me.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -1030,6 +1047,30 @@ class MainDataSource(
...
+     * if it's a server replay older than what we already have for this queue
+     * id. See [QueueInfo.isStaleReplay] for the comparison rules.
+     *
+     * `_queueInfos` is the single source of truth: server-event handlers
+     * write to it on admit, and [LocalPlayerRepository.onOptimisticQueueChange]
+     * mirrors optimistic-bump stamps into it via the callback wired in [init].
+     * That removes the need for a cross-store merge — the gate only has to
+     * look in one place.
+     */
+    private fun gateOrSkip(data: QueueInfo, label: String): Boolean {
```
formatBCE: Make this method extension to queue info e.g. `QueueInfo.takeIfNotStale(logLabel: String): QueueInfo?`. Return nullable QueueInfo here - it allows doing:
```
val data = event.queue().takeIfNotStale("QueueAdded") ?: return @collect
```
Beauty, isn't it? :) I love Kotlin for this.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -1476,6 +1577,13 @@ class MainDataSource(
                         }
 
                         is QueueTimeUpdatedEvent -> {
+                            // NOT staleness-gated. The event payload is just
```
formatBCE: I think all these comments are redundant, if code is clear enough...
If it's not - let's change the code! :)

formatBCE (changes requested):

formatBCE:
Sorry if there's too much. I feel like we need to align our agents. This change is good, but the code is not that good. :(

teancom:
Dude, absolutely don't worry about pushing back on bad code. Especially Kotlin I am *incredibly* new and miss stuff that I am at least a little more apt to catch with python or rust.

formatBCE (approved):

Issue #285, opened 2026-04-27
Allow transferring queue for current player anywhere in the app


seadowg:
We discussed this being one of the [design goals](https://github.com/music-assistant/mobile-app#design-goals), and it's currently the only one on the list that's missing from the app. 

On compact screens, the only obvious place for this to live would be on the left of the player name. That space is currently occupied by DSP settings in the expanded layout, however.

A few different options I think we should consider:

~~1. Removing DSP settings for every player and just linking to it from settings for the local player. Place Transfer on the other side of player from group.~~
~~2. Introduce an overflow menu with DSP settings in "Players" for the current player and remove the reorder anchor. We could add a way to switch to reordering mode (with the anchor) via long press or a specific button in the dialog. Again, we could then place Transfer on the other side of player from group.~~
~~3. Use a [split button](https://m3.material.io/components/split-button/overview) for the player button with Group, Transfer and DSP settings in the menu and remove the buttons for these actions.~~
4. Add a top right options menu to the compact and expanded player with Group, Transfer and DSP settings and remove the buttons for these actions. This menu could also include the "Go to artist" and "Go to album actions" discussed in https://github.com/music-assistant/mobile-app/issues/198.
~~5. Be more liberal about "anywhere in the app" and only show Group, Transfer and DSP settings in the expanded player view (as buttons to the side of the player name). This would free up room to make the compact player smaller.~~

formatBCE:
I think it's time for side 3-dot menu.
The question is - only in expanded state, or on mini-player too. 
I vote everywhere.

seadowg:
> I think it's time for side 3-dot menu.
The question is - only in expanded state, or on mini-player too.
I vote everywhere.

So option 4 yeah? I agree that in both places would be good. I like that we're striving to keep these actions available from as many places as possible, and it feels like a top right options menu wouldn't look too out of place in both the compact and expanded players. Also, very nice for it to be consistent between the two modes.

formatBCE:
> 
> So option 4 yeah? I agree that in both places would be good. I like that we're striving to keep these actions available from as many places as possible, and it feels like a top right options menu wouldn't look too out of place in both the compact and expanded players. Also, very nice for it to be consistent between the two modes.

Yes on all points. 
Thank you for raising this questions as we go! It looks like we're evolving in right direction!

seadowg:
@formatBCE feel free to assign this to me and I can do it while working on #198. Might as well add all these items into the overflow in one go!

[...]

PR #314, opened 2026-05-01
Add go to album/artist to expanded player

### LINKED ISSUE #198: It's hard to perform navigation between related items

seadowg:
Closes #198

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/ui/compose/common/items/ItemNavigationOptionsTest.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/ui/compose/common/items/ItemNavigationOptionsTest.kt
@@ -8,6 +8,7 @@ import io.music_assistant.client.data.model.client.AppMediaItemFixtures
 import io.music_assistant.client.support.get
 import io.music_assistant.client.ui.compose.common.OverflowMenu
 import musicassistantclient.composeapp.generated.resources.Res
+import musicassistantclient.composeapp.generated.resources.action_go_to_album
 import musicassistantclient.composeapp.generated.resources.action_go_to_artist
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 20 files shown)


androidApp/src/test/kotlin/io/music_assistant/client/ui/compose/common/items/ItemNavigationOptionsTest.kt
```diff
@@ -0,0 +1,49 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class ItemNavigationOptionsTest {
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun `AppMediaItem#navigationOptions does not include go to artist for track without artists`() {
+        val track = AppMediaItemFixtures.track(artists = emptyList())
+        composeTestRule.setContent {
```
seadowg, line 25: It seems like there's not really a good way to test `@Composable` helper functions - you could put assertions in `setContent`, but it all ends up feeling a bit weird. Even though they are a huge part of the Composable "standard library" (`rememeber`, `produceState` etc.), they don't get acknowledged a lot. I've found it's best just to "integration" test them within a very basic UI.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/settings/SettingsScreen.kt
```diff
@@ -73,8 +73,61 @@ import io.music_assistant.client.utils.SessionState
 import io.music_assistant.client.utils.isIpPort
 import io.music_assistant.client.utils.isValidHost
 import io.music_assistant.client.webrtc.model.RemoteId
-import musicassistantclient.composeapp.generated.resources.*
```
formatBCE, line 76: Do we do this?

seadowg: The expansion of imports? Looks like I have Android Studio set to use single name imports (Editor > Code style) so this is what happens on optimizing imports. I **think** that's the default setting, but it doesn't look like the [Kotlin conventions](https://kotlinlang.org/docs/coding-conventions.html) have a preference.

seadowg:
@formatBCE I've only added the overflow menu to the expanded player here. I'll experiment with it in both the compact and expanded with these actions and the player actions for #285.

formatBCE (approved):

PR #316, opened 2026-05-02
feat(observability): lifecycle markers + cancel-leak fix

teancom:
I added a bunch of logging to track down errors over the past few days. These are the ones that I thought were worth shipping to help with issue triaging in the future, along with fixing a small leak in sendRequest.

Robot:

Adds diagnostic logging across KtorServiceClient and AuthenticationManager to make stuck-reconnect / stuck-CarPlay sessions traceable from log captures.

- stateLabel / dcsLabel for compact, grep-stable session-state tags
- Transport state-edge logs (Connected / Reconnecting / Failed / Disconnected)
- isReadyForCommands edge collector
- Explicit "no recovery taken" branch in onAppForeground for the actually-stuck case (state != Connected, no savedInfo); happy path stays silent so logs aren't polluted with non-events
- "Reconnect not applicable" branch in onExternalConsumerActive
- sendRequest[msgId] markers: start/sent/resumed at .d (high volume), cancelled/transport=null at .i (rare, informational)
- AuthMgr per-branch trace via Logger.withTag("AuthMgr")

Bug fix: sendRequest now installs invokeOnCancellation to remove the rpcEngine callback if the parent coroutine is cancelled (e.g. a caller's withTimeoutOrNull fires). Without this, never-arriving responses leak orphaned callbacks for the lifetime of the session.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
@@ -140,7 +140,7 @@ class KtorServiceClient(private val settings: SettingsRepository) : ServiceClien
     override fun onAppBackground() {
         isInBackground = true
         backgroundedAt = currentTimeMillis()
-        logger.i { "App backgrounded" }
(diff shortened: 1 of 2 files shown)


formatBCE (approved):
Looks good - let's see how it runs ;)

PR #318, opened 2026-05-02
fix(androidApp, composeApp): Disconnect sendspin client on close when not playing audio.

### LINKED ISSUE #302: Pausing playback causes client to remain in background indefinitely
MA Version: 2.8.6
Connection Method: Direct Connection
Client: Android (Samsung)
Client Version: 2026.04.28.1810

Issue: When playback is paused the app will indefinitely remain in the background. When playback resumes after an extended time period, a force stop is required for audio to play (UI features function as intended, and track progress is still displayed accurately prior). This only occurs if there is a queue on the sendspin client.

Steps to reproduce:
1. Begin playback
2. Pause playback
3. Close the app
4. Wait an extended time period (normally 10-15 minutes will do)
5. Resume playback

Blu3sClues:
Created a function for when the app is closed, which checks for if music is playing and disconnects the sendspin client if nothing is playing.

When an android user removes the app from their recent apps, the function is triggered and checks for active playback. If no playback is active, the sendspin client is disconnected and the application stops itself and removes from the foreground.

If playback is active, it continues playing in the background to prevent accidental closures.

Fixes #302

### DIFF
--- a/androidApp/src/main/kotlin/io/music_assistant/client/services/MainMediaPlaybackService.kt
+++ b/androidApp/src/main/kotlin/io/music_assistant/client/services/MainMediaPlaybackService.kt
@@ -303,6 +303,18 @@ class MainMediaPlaybackService : MediaBrowserServiceCompat() {
         dataSource.selectPlayer(list[newIndex].player)
     }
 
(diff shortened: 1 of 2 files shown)


formatBCE (approved):
Great! Thank you!

PR #319, opened 2026-05-02
CarPlay reliability

teancom:
This is the PR with the actual CarPlay reliability fixes that I've been working on. Specifically, we had some forever-hangs that now get timeouts, we add a 'Disconnected' row to the UI if we _do_ get timeouts, and we can now drill down into the actual albums of an artist, etc.

- KmpHelper: nullable-on-timeout fetcher contract (5s withTimeoutOrNull), Cancellable + observeReadiness/observeLocalPlayerPresence subscriptions, drilldown fetchers (artist→albums, album→tracks, playlist→tracks).
- CarPlay scene: readiness-gated taps with idempotent offline alert, generation counter on the recommendations fetch, type-aware drilldown dispatch, popToRoot before pushing Now Playing to dodge the depth-5 hierarchy crash, one-shot initial Now Playing push on local-player presence.
- CarPlayContentManager: drilldown wrappers + mapItems threading.
- SiriIntentHandler: nil → empty coalesce on the four search call sites.
- Lifecycle markers via OSLog at .default for post-drive sysdiagnose.

### DIFF
--- a/composeApp/src/iosMain/kotlin/io/music_assistant/client/di/KmpHelper.kt
+++ b/composeApp/src/iosMain/kotlin/io/music_assistant/client/di/KmpHelper.kt
@@ -1,5 +1,6 @@
 package io.music_assistant.client.di
 
+import co.touchlab.kermit.Logger
 import io.music_assistant.client.api.Request
 import io.music_assistant.client.api.ServiceClient
 import io.music_assistant.client.auth.AuthenticationManager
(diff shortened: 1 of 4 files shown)


teancom:
I've re-run the Android tests a couple of times. I think (since I didn't touch any Android code) that it's probably not something I did? But if I'm wrong, I'm happy to fix it.

formatBCE (approved):

teancom:
I *think* I found the cause of the android test failures: https://github.com/music-assistant/mobile-app/pull/321

PR #321, opened 2026-05-02
fix(test): wait for track-name update in assertPlayer

teancom:
A couple tests added in #314 would only intermittently pass (almost always locally, very flaky in CI). This should fix that, if my robot sidekick is right (it worked well locally, but so did the first version).

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/support/pages/SharedActions.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/support/pages/SharedActions.kt
@@ -113,12 +113,19 @@ fun <T : ComposePage> T.assertPlayer(
         }
     }
 
+    // waitUntil — the track-name flow update can lag the play action.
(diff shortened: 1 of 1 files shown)


formatBCE (approved):

seadowg:
@teancom looks good! We'll most likely want to prefer "wait" assertions over immediate ones as `ComposeTestRule` is only synchronized with the UI thread ([details here](https://developer.android.com/develop/ui/compose/testing/synchronization) if you're interested) and the app has a lot of event reaction logic that uses `Flow` either gets executed on a background thread or scheduled on the UI thread (giving the tests a window to assert before things have actually completed).

A nice trick if you spot a flake like this is to stick a Thread.sleep(1000) in before the flaky assertion, see if it passes every time, and if so then replace with a `waitUntil` or one of the async assertions like `waitUntilAtLeastOneExists`.

Issue #278, opened 2026-04-26
UI - Quality shown is always the output on device, not the input quality (so 320 source may be shown as alac/flac 44.1/16 instead of eg mp3)


remon1496:
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/25f4b5f2-cf43-42c0-8d98-149c3108ec75" />

The app always shows the output settings in the quality overview. This is technically correct, but I think the quality should be based on the lowest quality bottleneck just as in web ui. If source is lower than output, show source quality, is output quality is lower quaility than source, then show output limit. Maybe just show the whole stack.

Lower priority for now, but audiofiles will most def. not like current behaviour/feedback.

<img width="692" height="448" alt="Image" src="https://github.com/user-attachments/assets/8f7ac29d-6d9b-43a8-86db-4d9f8dd61bec" />

seadowg:
I think this should be low priority for the moment, but I reckon a signal path dialog/bottom sheet with a similar UI to the screenshot, but that also links DSP settings would be great (it'd be cool for web/desktop to also link DSP from signal path as well). This could be opened with a little feel good quality light button like in MA web or Roon as well.

remon1496:
Agreed! That would be awesome. :)

PR #327, opened 2026-05-03
Move queue actions and DSP into overflow menu

### LINKED ISSUE #276: UI/UX - Spacing control buttons in expanded player screen
Testflight 1.0 (10)

image

The player controls on the expanded player are a bit too close to one another. Already had a view instances where I accidentally enabled repeat instead of moving to the next song. Same for restarting track (initiating accidental shuffle)

### LINKED ISSUE #285: Allow transferring queue for current player anywhere in the app

seadowg:
Closes #285
Closes #276

Screenshot_20260503_140653

I've also removed the skip backwards button from the compact player to increase the spacing available for other elements ([we'd discussed that button being less important](https://github.com/music-assistant/mobile-app#design-goals)):

Screenshot 2026-05-04 at 12 02 47

Just a note on testing: I haven't added any as I've mostly just moved things around that were already untested (and require bootstrapping to write tests for). Although that can of course also break things, I was happier to just get this in, and then I can follow up with backfilling tests for these features.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -145,6 +145,7 @@
     <string name="queue_no_items">No items</string>
     <string name="queue_not_loaded">Not loaded</string>
     <string name="queue_cannot_play">Cannot play this item</string>
+    <string name="queue_clear">Clear queue</string>
 
     <!-- Players -->
     <string name="players_title">Players</string>
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -59,7 +59,6 @@
     <string name="cd_fully_played">Fully played</string>
     <string name="cd_in_progress">In progress</string>
     <string name="cd_provider_icon">Provider Icon</string>
-    <string name="cd_dsp_settings">DSP Settings</string>
     <string name="cd_playing">Playing %1$s</string>
     <string name="cd_current_player">Current player: %1$s</string>
 
@@ -152,6 +151,7 @@
     <string name="players_group_settings">Group settings</string>
     <string name="players_group_volume">%1$s group volume</string>
     <string name="players_nothing">Nothing playing</string>
+    <string name="players_dsp_settings">DSP settings</string>
 
     <!-- DSP -->
     <string name="dsp_title">DSP settings</string>
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -36,7 +36,6 @@
     <string name="common_delete">Delete</string>
     <string name="common_create">Create</string>
     <string name="common_done">Done</string>
-    <string name="common_transfer">Transfer</string>
 
     <!-- Content descriptions -->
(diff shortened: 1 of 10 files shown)


seadowg:
@formatBCE worth taking a look to see how you feel about the split button here! I'm liking it. I think the spacing in compact mode is a little tight, but that's true of the previous layout as well, and I'll try tweaking that before marking this ready for review.

I do think it would work even better if we adopt the same group "labelling" as desktop here and basically the player name update when grouped (above would be "nexus + 1" for example) and switch the group button to the "link" icon used on desktop as well (keeping the outlined/primary style switching based on if the player is grouped or not):

Screenshot 2026-05-03 at 14 10 45

This would give people consistent language even though the interface is a little different. The code deals with a case for a grouped player (`player.player.isGroup`) which I don't have an example of myself to see what it looks like on desktop.

remon1496:
Screenshot looks slick!

formatBCE:
We had that in-title number before. Can bring it back. Or have a badge with number on the button, like web does on players list.
P.S. I just realized that you use light theme.
Good that we have different preferences. :)

seadowg:
> P.S. I just realized that you use light theme.
Good that we have different preferences. :)

Only in the emulator 😆. Personal device is also dark theme.

> We had that in-title number before. Can bring it back. Or have a badge with number on the button, like web does on players list.

So the player name for a player "My player" would be:

- "My player" if it's not part of a group
- "My player + 1" if it's part of a group with 1 other speaker (for example)

And then a group "My group" with 2 items (again, for example) could be "My group (2)"?

formatBCE:
Yeah it was like this before. I kinda like it current way - but if you think that name suffix is better - I'm fine.

seadowg:
@formatBCE I think we should do some more general tweaking of the expanded player layout, but I think that can all continue to evolve as part of #113 (to get things working for larger screens and probably rework the whole layout) and #278 (we'll get the player type and format off the screen). Right now for #276, I've just focused on making sure there's enough space between things to reduce accidental taps.

formatBCE (approved):

PR #338, opened 2026-05-06
Fix transfer autoplay

### LINKED ISSUE #287: Bug? - Transferring queue doesn't auto start playback
Testfight 1.0 (11)

Using MA web, when you transfer a queue, it automatically starts playback on the player/group the queue is transferring to. On the mobile app, you have to manually start the playback after transferring. I think we would want to use the same behaviour in the app as on MA web?

seadowg:
Closes #287

I'd originally just been backfilling tests, but then discovered that I'd accidentally written a failing test for a bug. I was also going to write tests for the actual queue items themselves, but figured I should merge early to fix the autoplay issue.

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/feature/QueueTest.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/feature/QueueTest.kt
@@ -35,12 +35,29 @@ class QueueTest {
         serviceClient.addToLibrary(track)
 
         val player = ServerPlayerFixtures.player()
-        serviceClient.addPlayer(player)
+        serviceClient.addPlayers(player)
 
         launchLoggedInApp(composeTestRule, serviceClient)
             .clickOnMedia(album)
             .clickPlay()
             .expandPlayer(player.displayName, playing = true, item = track.name)
             .clearQueue()
+    }
+
+    @Test
+    fun `can transfer queue to another player`() {
+        val album = ServerMediaItemFixtures.album()
+        val track = ServerMediaItemFixtures.track(album = album)
+        serviceClient.addToLibrary(track)
+
+        val player1 = ServerPlayerFixtures.player()
+        val player2 = ServerPlayerFixtures.player()
+        serviceClient.addPlayers(player1, player2)
+
+        launchLoggedInApp(composeTestRule, serviceClient)
+            .clickOnMedia(album)
+            .clickPlay()
+            .expandPlayer(player1.displayName, playing = true, item = track.name)
+            .transferQueue(player2.displayName)
     }
 }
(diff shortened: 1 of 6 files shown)


androidApp/src/test/kotlin/io/music_assistant/client/support/pages/ExpandedPlayerPage.kt
formatBCE, line 1: Geez this stuff becomes extensive as hell! We will be writing tests for these tests soon. :)

androidApp/src/test/kotlin/io/music_assistant/client/feature/QueueTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+        val player1 = ServerPlayerFixtures.player()
+        val player2 = ServerPlayerFixtures.player()
+        serviceClient.addPlayers(player1, player2)
+
+        launchLoggedInApp(composeTestRule, serviceClient)
+            .clickOnMedia(album)
+            .clickPlay()
+            .expandPlayer(player1.displayName, playing = true, item = track.name)
+            .transferQueue(player2.displayName)
```
formatBCE, line 61: No assertion? I mean, it will fail if anything there isn't accessible - but the failure reason will be hidden in the implementation chain... 
Don't you think that tests aren't "unit" tests anymore?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
```diff
@@ -553,7 +553,7 @@ private fun PlayerOverflowMenu(
                                 QueueAction.Transfer(
                                     queueId,
                                     playerData.player.id,
-                                    playerData.player.isPlaying,
+                                    currentPlayer.player.isPlaying,
```
formatBCE, line 556: Hmm, when did that break... I'm sure it was working at the very beginning... Thanks for surgical fix!

formatBCE (approved):

androidApp/src/test/kotlin/io/music_assistant/client/feature/QueueTest.kt
seadowg: > No assertion? I mean, it will fail if anything there isn't accessible - but the failure reason will be hidden in the implementation chain...

`transferQueue` ends with an assertion:

```kotlin
return ExpandedPlayerPage(player, true, item, composeTestRule).assertOnPage()
```

There's a trade-off with having the failure happen at a line in the test here by having page objects, but in return the tests should be easier to write, change and read.

> Don't you think that tests aren't "unit" tests anymore?

I wouldn't consider these tests unit tests! The only thing being faked/mocked/stubbed here is the `ServiceClient` so at this level we're testing pretty much the full app. For this codebase, I'd consider these tests as "end-to-end" (or as much end-to-end as you'd want anyway).

formatBCE: Okay. :)

PR #342, opened 2026-05-07
Fixed music delay when was in background not playing.

formatBCE:
Bug: pause player, send app to background (or just disconnect from Android Auto), and then in a while try playing music.
Outcome: the playback is shifted in time for the time the phone was in "doze" - looks like it's not playing at all, if it was off for too long.
This is fix for that.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
@@ -22,6 +22,7 @@ interface ServiceClient {
 
     fun onAppForeground()
     fun onAppBackground()
+    val foregroundEvents: Flow<Unit>
     fun disconnectByUser()
     fun connect(connection: ConnectionInfo)
     fun connectWebRTC(remoteId: RemoteId)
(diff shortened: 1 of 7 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
```diff
@@ -22,6 +22,7 @@ interface ServiceClient {
 
     fun onAppForeground()
     fun onAppBackground()
+    val foregroundEvents: Flow<Unit>
```
seadowg, line 25: It looks like you just need to override this in `FakeServiceClient` (can just be set to a `MutableSharedFlow`) to get tests working. 

Just a thought: instead of needing to host a new event in `ServiceClient`, could we just make `MainDataSoruce` also lifecycle aware? It could either implement `LifecycleObserver` or also get notified from `AppLifecycleObserver` like `ServiceClient`. That feels like it would isolate this feature(/bug fix) a little more.

formatBCE: Yeah I just had no time for this test fix.
I'll think about your proposal. I had some arguments yesterday, don't remember what exactly.

formatBCE: Tried to make MainDataSource lifecycle-aware. It imposed 2 problems:
1. Too convoluted for single use case.
2. MainDataSource already "has-a" service client. It owns that object too. So it would be either hoisting whole functionality (which adds to already bloated MainDataSource even more), or keep it there in ServiceClient... There's 3rd way - extract it. It would be big refactoring - so I'll save it to the future.

PR #351, opened 2026-05-08
Don't display version twice on item details page

seadowg:
We display item versions underneath the name, so we don't need to add the suffix.


seadowg:
Looks like I've broken some iOS code! A good opportunity to dive into some Swift.

formatBCE:
It's easy fix :) replace title with name where it asks...

seadowg:
> It's easy fix :) replace title with name where it asks...

Yeah, looks like it! I'll get my local iOS build working, and then make the changes in XCode. Will be good to be setup for that!

formatBCE:
Total overkill, but good that you eager to dive into that :)

seadowg:
> Total overkill

haha agreed! But I'd had it on the to-do list anyway, and this isn't a particularly pressing fix.

seadowg:
@formatBCE actually already had enough of the build working to get the correct compilation failures! Will save a full setup for another day.

formatBCE:
> @formatBCE actually already had enough of the build working to get the correct compilation failures! Will save a full setup for another day.

Save this windy road for later! :)

formatBCE (approved):

PR #354, opened 2026-05-08
Update README with in-car design goals

remon1496:
Added information about Apple CarPlay and Android Auto design goals.


formatBCE (approved):

remon1496:
Whoops, removed that one a bit too early.

PR #384, opened 2026-05-13
Add dismiss on swipe down.

formatBCE:


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
```diff
@@ -307,8 +315,62 @@ private fun ExpandedPlayerPage(
     livePositionFlow: Flow<Double>?,
 ) {
     val isLargeScreen = WindowClass.isAtLeastLarge()
+    val dismissThresholdPx = with(LocalDensity.current) { 120.dp.toPx() }
+    val dismissNestedScroll = remember(onClose, dismissThresholdPx) {
+        object : NestedScrollConnection {
```
seadowg, line 320: We might want to pull this all out into our `FloatingBar` custom component so it's just part of the "expanded" mode implementation.

seadowg (commented):
Wow this is great! I think there's some sight weirdness that the animation and movement doesn't suggest that the component physically moves up or down (it "expands"), but on iOS this is still going to be the most natural way to close it given that it's a "modal" presentation. I think a lot of people will naturally do this on Android as well.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayersPager.kt
formatBCE: Yeah, I guess that would be better!

PR #395, opened 2026-05-15
Fix backstack weirdness when restoring process

seadowg:
This fixes issues when the process is recreated and the "original" items of a back stack were treated as the ones in place before the restore causing the "return" action (clicking the nav item when already on it) to act weirdly.

Now when restoring, the back stacks should remain the same as they were before and "return" should work as expected. One caveat that was already the case, when restoring the app will always switch the "Home" stack because the "current stack" state doesn't survive process recreation (it's just a `State` wrapped in a `remember`). If that's something we see a lot in the wild, we can improve on it.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/MultiBackStack.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/MultiBackStack.kt
@@ -18,7 +18,7 @@ import androidx.navigation3.runtime.NavKey
 class MultiBackStack<T : NavKey>(private val backStacks: List<MutableList<T>>) {
     var currentBackStack by mutableStateOf(0)
 
-    private val originalItems = backStacks.map { it.toList() }
+    private val roots = backStacks.map { it.first() }
 
     /**
      * Clear the current back stack and reset it back to its original state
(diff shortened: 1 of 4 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/MultiBackStack.kt
```diff
@@ -63,23 +58,15 @@ class MultiBackStack(private val backStacks: List<NavBackStack<NavKey>>) {
...
 
-    @Composable
     fun toEntries(entryProvider: (NavKey) -> NavEntry<NavKey>): List<NavEntry<NavKey>> {
         val activeBackStacks = if (currentBackStack == 0) {
             listOf(backStacks[0])
         } else {
             listOf(backStacks[0]) + listOf(backStacks[currentBackStack])
         }
 
-        val saveableStateHolderForHome = rememberSaveableStateHolder()
```
seadowg, line 74: I don't think all this saved state code is helping anything. The back stacks themselves are saved/remembered and that's where the actual "state" here lives. This process is just converting that state into a form that `NavDisplay` can use.

formatBCE (approved):
Love simplification!

PR #418, opened 2026-05-19
Convert `ActionViewModel` data classes into interfaces

seadowg:
Calls back up the `ActionViewModel` were being propagated via data classes holding callbacks which created a lot of noise in the code. I've switched this up so that there is a series of interfaces (corresponding to the previously existing data classes). This means that there is less noise when passing params down to Composables, but also keeps the actual detail of the view model away from the pure view code (as it was before).

I've also made the "screen" level Composables consistent so that they all accept ViewModels as they're main input. The actual creation of ViewModel classes all happens in the navigation roots where it's (hopefully) clearer which lifecycle/scope we're operating in.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
@@ -68,6 +68,7 @@ import io.music_assistant.client.ui.compose.common.ToastHost
 import io.music_assistant.client.ui.compose.common.ToastState
 import io.music_assistant.client.ui.compose.common.items.AlbumWithMenu
 import io.music_assistant.client.ui.compose.common.items.ArtistWithMenu
+import io.music_assistant.client.ui.compose.common.items.LibraryActions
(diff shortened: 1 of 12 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -153,20 +156,38 @@ fun ItemDetails(
     onAlbumsSortChanged: (SubItemContext, SortOption) -> Unit = { _, _ -> },
     onPlayableItemsSortChanged: (SubItemContext, SortOption) -> Unit = { _, _ -> },
 ) {
-    val playlistActions = ActionsViewModel.PlaylistActions(
-        onLoadPlaylists = geEditablePlaylists,
-        onAddToPlaylist = addToPlaylist,
-    )
+    val playlistActions = object : PlaylistActions {
```
seadowg, line 159: This is messy, but it was before. I was trying to keep myself from descending a rabbit hole!

formatBCE (approved):

formatBCE:
Very nice!

Issue #346, opened 2026-05-07
Notify user on unrecoverable errors, currently they are silent in UI (but shown in log)


remon1496:
Radio Mode currently fails silently when only a Local file system is setup. Radio Mode/Dont stop the music! requires a music provider to have the abbility to queue similair items.

Webui shows error (found out while testing MA beta).

Should we hide radio mode when no compatible provider is setup, or simply notify that its failing?

Might be that in later releases core function for this changes because of new audio analysis functions.

<img width="384" height="94" alt="Image" src="https://github.com/user-attachments/assets/9b37228c-d78a-4fed-99ce-3adac701add2" />

Blu3sClues:
Are you referring to the "Start Radio" option being hidden? Either way, I tested this and the logs do show this:

"Error/RpcEngine: RPC error for message [REDACTED_ID]: Answer(json={"message_id":"[REDACTED_ID]","error_code":9,"details":"No Music Provider found that supports requesting similar tracks."})

I think it would be a good idea to implement some kind of visual message for these RPC errors. As you mentioned, they currently are silent and just pop up in the logs. This would allow us to display this error and any other errors that are sent, without hard coding an error ourselves in case of future changes.

remon1496:
> Are you referring to the "Start Radio" option being hidden? Either way, I tested this and the logs do show this:
> 
> "Error/RpcEngine: RPC error for message [REDACTED_ID]: Answer(json={"message_id":"[REDACTED_ID]","error_code":9,"details":"No Music Provider found that supports requesting similar tracks."})
> 
> I think it would be a good idea to implement some kind of visual message for these RPC errors. As you mentioned, they currently are silent and just pop up in the logs. This would allow us to display this error and any other errors that are sent, without hard coding an error ourselves in case of future changes.

Yeah, also agree on just forwarding them in a visual manner is the way to go. Keep maintenance on client app as easy as possible. :)

formatBCE:
Yup, I was thinking on basic toast messages service that would show these unrecoverable errors to user.

Blu3sClues:
Yup, I agree that a basic toast is the way to go.

PR #421, opened 2026-05-20
Dont stop the music

### LINKED ISSUE #385: Feature - Don't Stop The Music! from 3-dot menu in expanded player
## Summary
MA app has an options to enable Radio mode, this function lives in different places. Currently this mode can be activated from all these places, but not from the Player Screen when something is playing already. It would be nice to quickly enable 'Don't stop the music!' via the 3-dot submenu on the expanded player screen.

### The current UX issue that exists without this option
Because 'Radio' mode currently can only be activated when starting to play something, you have to go back to an album/artist/track and enable radio mode via this place. This replaces the queue (stop playback and start playing the new queue).

### How to solve it
This UX would be more streamlined when being able to enable 'Don't stop the music!' (which leverages radio mode), from the 3-dot menu in the expanded player. The music would keep playing, keeps the queue and adds related items to bottom of the current queue. It will utilize everything in current session history (played queue items) to add new related items to bottom of the queue.

### The action and it's logic
The following would happen when this button is clicked, as extracted from the docs that live here > https://www.music-assistant.io/usage/

"_The Don’t Stop The Music (DSTM) option can be enabled if a source is available which supports dynamic tracks (i.e. Apple, Deezer, Spotify, Subsonic, Tidal and YTM). When DSTM is on, radio mode will be automatically enabled when the last track of the queue is reached and if any dynamic tracks can be resolved from one of the sources. The added tracks will be based on the played items in the queue._"

formatBCE:
Fixes #385 and #346


remon1496:
Im missing the: 'Enable Don't Stop The Music' and 'Disable Don't Stop The Music' strings from the strings XML (as far as I understand the code). Are these currently hardcoded?

formatBCE:
> Im missing the: 'Enable Don't Stop The Music' and 'Disable Don't Stop The Music' strings from the strings XML (as far as I understand the code). Are these currently hardcoded?

You bet they are :(

PR #429, opened 2026-05-21
Fix landscape player on devices in between expanded and large screen classes

### LINKED ISSUE #383: Landscape view not splitting on iPad
## Summury
PR https://github.com/music-assistant/mobile-app/pull/372 and https://github.com/music-assistant/mobile-app/pull/368 added UI tweaks and Landscape layout for devices with larger screens. This caused UI isseus in the Expanded Player view on iPadOS devices in landscape mode.

## Devices and app-version
- iPad A16 - TestFlight 1.0 (25)

## Observed on iPad
Expected similar view is shown in: https://github.com/music-assistant/mobile-app/pull/368
The queue isn't put into split view as expected in example from PR.

Image
Image

seadowg:
Closes https://github.com/music-assistant/mobile-app/issues/383

We could do more tweaking in this area, but for the moment I'm just fixing this and keeping the point where we go "landscape" as the large size class.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
@@ -50,8 +50,8 @@ fun String.isIpPort(): Boolean {
 
 fun Modifier.conditional(
     condition: Boolean,
-    ifTrue: Modifier.() -> Modifier,
     ifFalse: (Modifier.() -> Modifier)? = null,
+    ifTrue: Modifier.() -> Modifier,
 ): Modifier {
     return if (condition) {
         then(ifTrue(Modifier))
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
```diff
@@ -50,8 +50,8 @@ fun String.isIpPort(): Boolean {
 
 fun Modifier.conditional(
     condition: Boolean,
-    ifTrue: Modifier.() -> Modifier,
```
formatBCE, line 53: Why? :)

formatBCE: Ah, to have inline lambda for true! Got it.

formatBCE (approved):

PR #430, opened 2026-05-21
Show quality badge and per-item quality info

### LINKED ISSUE #278: UI - Quality shown is always the output on device, not the input quality (so 320 source may be shown as alac/flac 44.1/16 instead of eg mp3)

formatBCE:
Fixes #278

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/AudioChainDialog.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/AudioChainDialog.kt
@@ -1,0 +1,166 @@
+@file:Suppress("MagicNumber")
+
+package io.music_assistant.client.ui.compose.home.players
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Card
+import androidx.compose.material3.HorizontalDivider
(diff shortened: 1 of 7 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/AudioChainDialog.kt
```diff
@@ -0,0 +1,166 @@
...
+        player.parentBind?.let { put(it.id, it.name) }
+    }
+    Dialog(onDismissRequest = onDismissRequest) {
+        Card(
+            modifier = Modifier.fillMaxWidth(),
+            shape = RoundedCornerShape(16.dp),
+        ) {
+            Column(modifier = Modifier.padding(20.dp)) {
+                Text(
+                    text = "Audio chain",
```
remon1496, line 54: Hardcoded? :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/AudioChainDialog.kt
```diff
@@ -0,0 +1,166 @@
...
+                    style = MaterialTheme.typography.headlineSmall,
+                )
+                Spacer(modifier = Modifier.height(12.dp))
+
+                ChainStage(
+                    header = "Input",
```
remon1496, line 60: Also missing from translatable strings.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/AudioChainDialog.kt
```diff
@@ -0,0 +1,166 @@
...
+                    title = queueTrack.provider?.replaceFirstChar { it.uppercaseChar() },
+                    format = queueTrack.format,
+                )
+
+                queueTrack.dsp.orEmpty().forEach { (playerId, dspSettings) ->
+                    HorizontalDivider(modifier = Modifier.padding(vertical = 8.dp))
+                    val playerName = playerNames[playerId] ?: playerId
+                    ChainStage(
+                        header = "Output",
```
remon1496, line 69: And this one also. :)

PR #445, opened 2026-05-24
Add new design for library

### LINKED ISSUE #337: Library redesign ideas
**Current situation:**
The library lives on a single 'root' screen with a horizontal (scrollable) navigation bar, which can be used to navigate between the different library types (albums, tracks, playlists etc). It's kind of hard to navigate the different 'sub libraries). Depending on the screen width you have to horizontal scroll to access items that fall outside of the current viewport. This is especially hard during one handed use, where you have to reach out with your thumb to the top of the sceen to horizontal scroll where you want to go.

**Screenshot of the current 'root' library:**
Image

**The proposal:**
Replace current rootview with a intermediate screen 'root' screen where all 'child libraries' are listed (list/grid view).

**The root library view from which you can navigate to child libraries:**
Image

After navigating to a 'child library' (eg. artists) you land in the artists view of the library. Depending on the library type this is populated with:

**In the top row:**
- Back button
- Child Library name
- Grid/List switch - Save the set grid/list view preference per current view, this enables eg. default grid view for artists and default list view for albums/tracks/playlist, or whatever you prefer) - Might also be put in 3-dot menu with more (future) settings/options?

Below that, visible before scrolling down, hidden after scrolling down and re-appear when scrolling back up:
- Quick search bar
- Filter favorite button
- Sort options

Image

The individual items page (eg. artist) stays the way it is for this idea, but can ofc. be tweaked later.

Image

When quick searching in the local (child) library, it might happen that the thing you search for isn't in the local library. It would make sense to show the 'try global search' button in this case (same as in webui). Clicking the button will redirect to the 'global search page' with the current search library term and fill this in the 'global search bar'. The filter of content type could also be pre-set based on the reffering 'child library' page. (Eg.: when searching for artist 'fm-84 doesnt gave any library results, pre-fill 'fm-84' in global search bar and pre select the 'artists' filter.

Image

A view of the Playlists child library:
Image

seadowg:
Closes #337

Todo: 

- [x] Restore customization of "tabs" (needed at least for Auto)
- [x] Clean up `LibraryTabsViewModel`
- [ ] ~~Add global search navigation on empty results~~ (will pull this out as a separate issue - #456)


remon1496:
Comments:

1. On main library screen, I feel text (eg. 'Artists') in library items could be closer to the icons. Allows for longer words in translation.
2. We def. need edit (show/hide/re-order) on main library screen.
3. Search bar in an individual library type has only a tiny bit of margin on top and bottom after scroll. Could the search bar be (a bit) less tall? Feels more balanced. Or nothing we can do here?
4. Font size of all search bars is very big, not specific to this pr. Lots of padding top/bottom of text (I don't think this is the line height?)
5. I don't mind not always seeing what order I've sorted my items in. Single tap away if I need to know. That tap is needed anyways if I need it to be set different.
6. I miss the 'favorite' filter though.

While it doesn't bother me personally, the lack of a visible 'back' cue when the search bar is open could make some users feel stuck — like they're on a random screen with no way out.
I think sacrificing a bit of vertical space for a small bottom sheet (housing the sort and favorite controls) would free up enough room in the top bar to always show the 'back' cue there.

So far, I haven't been able to break anything. Long queries, going (unrealistically deep) in the stack, and simply mindlessly tapping around.

seadowg:
> On main library screen, I feel text (eg. 'Artists') in library items could be closer to the icons. Allows for longer words in translation.

This is currently using standard Material spacing, so I'd rather not tweak it without input from a designer.

> We def. need edit (show/hide/re-order) on main library screen.

Done!

> Search bar in an individual library type has only a tiny bit of margin on top and bottom after scroll. Could the search bar be (a bit) less tall? Feels more balanced. Or nothing we can do here?
Font size of all search bars is very big, not specific to this pr. Lots of padding top/bottom of text (I don't think this is the line height?)

Both of these can be tweaked later, but again it means opting out of standard Material styling - we'd probably want to think carefully about changes as they'll look that way for a reason.

> I don't mind not always seeing what order I've sorted my items in. Single tap away if I need to know. That tap is needed anyways if I need it to be set different.

Agreed! My thinking is that you only really care about that value when it's not working for you, and you want to change it any way.

> I miss the 'favorite' filter though.

I'd really like to push to leave this out for now, but it can be reintroduced with a filter flow later (that could allow for other kinds of filtering). Just as some impromptu user research: what significance do items marked as favourite have for you here? We're within your library, so I'm interested to understand what extra value you get from toggling the heart on an item.

formatBCE:
Regarding favourites - it's really important thing. For every artist I have favourite songs. And the library often is convoluted with artists from collaborations etc., or it's common library for the whole family - but favs are per-user... Etc., etc.

In any way, favs are the functionality used by people, and easily accessible. Hiding it under other 2 clicks like in filter view would be a mistake.

The crusade for vertical space should finish somewhere. :) For me personally, even old header is okay, since it's hiding. If you're so eager to make more, let's hide bottom elements as well. But I find it enough.

seadowg:
OK, so given these comments and the ones at https://github.com/music-assistant/mobile-app/issues/337#issuecomment-4534384065, let's move back to having search and favourites below the app bar (with sort options showing) for now.

seadowg:
Ok that's the sort and favourite chips restored! Let me know what you think. Material doesn't currently have a concept for an app bar that has a second row of controls like this, so I've had to hack something together that we can probably hide behind a custom component that can hopefully get iterated on.

@remon1496 I'll send you a test APK again so you can have a play with it.

remon1496:
> I'll send you a test APK again so you can have a play with it.

Played around with it more and I like it! The back button cue 'missing' with search open, might still be a thing (there's room now with the sort moved to the bottom sheet, so it could be restored there if we get the feedback?).

I feel like moving the favourite and sort of bottom sheet was a good call.

The 'space lost' by reintroducing the bottom sheet could be regained by tweaking the spacing on the elements, but indeed would need help from a designer when moving away from standard elements and spacings.

Agian, I wasn't able to break anything on Android.

seadowg:
@formatBCE there are a few other things we might want to follow up with here:

- the "tabs" metaphor could probably be replaced with "library categories" throughout
- we could DRY up our search input implementations so that global and quick search share the same component
- tweaks to search style 

That said, it's probably worth merging this if there are no blockers as it gets us the new design in for feedback from more peeps.

remon1496:
Just to clarify, are all strings translatable or are they hardcoded in the new library? :)

seadowg:
> Just to clarify, are all strings translatable or are they hardcoded in the new library? :)

Good shout! The library category strings are hardcoded still, but that kind of relates to my earlier thought quite conveniently:

> the "tabs" metaphor could probably be replaced with "library categories" throughout

Could you file an issue about needing those to be translatable?

formatBCE (approved):

PR #448, opened 2026-05-25
Speed up iOS release link with smallBinary

teancom:
Locally on Apple Silicon, linkReleaseFrameworkIosArm64 drops from 5m18s to 3m49s and the static framework shrinks from 180MB to 158MB.

This is part of me trying to figure out how to make the iOS build and release process less painful, without paying for larger GitHub action runners.


formatBCE (approved):
Let's try it

PR #450, opened 2026-05-26
Fix iOS Ktor-Darwin uncaught WebSocket exception → SIGABRT

teancom:
I got a crash report from @remon1496 and tracked it down to this. I went with this (small) refactor approach vs. just catching the one exception so that we don't mask real problems / conflate them with network issues. And then found a couple of latent issues while I was in the neighborhood.

Ktor's Darwin engine can throw NSError-derived Kotlin exceptions from NSOperationQueue workers, outside our try/catch boundaries. With no CoroutineExceptionHandler in the failing coroutine's context, they reach Kotlin/Native's propagateExceptionFinalResort and abort the process.

Each transport now derives its own internal CoroutineScope from SupervisorJob(parentJob) + a state-aware CoroutineExceptionHandler: probe/forceReconnect when Connected; transition to Failed when Connecting/Reconnecting so the existing observer → Disconnected.Error → kickRecovery → reconnectFromHistory chain self-heals instead of getting permanently stuck. Adds Transport.close() (= scope.cancel()) called at every teardown site so we don't leak the old transport's SupervisorJob into the parent's job hierarchy on rebuild.

KtorServiceClient gets a log-only handler as a backstop for non- transport launches (settings collector, observe-transport callbacks, handleIncomingMessage's uncaught Event branch) not otherwise wrapped in try/catch.

While in the area:
- @Volatile on DirectTransport.session/messageCounter and SignalingClient.session — pre-existing races widened by routing handler execution onto the failing thread.
- WebRTCConnectionManager.handleSignalingError/handlePeerDisconnected now wrap their scope.launch { cleanup() } in try/catch.
- WebRTCTransport.onNetworkLost() also cancels connectionJob so a forceReconnect-in-flight can't race a parallel startReconnection().

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
@@ -30,6 +30,7 @@ import io.music_assistant.client.utils.myJson
 import io.music_assistant.client.utils.resultAs
 import io.music_assistant.client.utils.update
 import io.music_assistant.client.webrtc.model.RemoteId
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.IO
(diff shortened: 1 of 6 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
```diff
@@ -67,8 +68,15 @@ class KtorServiceClient(
     private val settings: SettingsRepository,
     private val errorBus: ErrorMessageBus,
 ) : ServiceClient, CoroutineScope, KoinComponent {
+    private val logger = Logger.withTag("ServiceClient")
     private val supervisorJob = SupervisorJob()
-    override val coroutineContext: CoroutineContext = supervisorJob + Dispatchers.IO
+
+    private val scopeExceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        logger.e(throwable) { "Uncaught exception in KtorServiceClient scope (non-transport)" }
```
formatBCE, line 75: Do we want to re-throw here? To allow catching downstream?

formatBCE (approved):

PR #460, opened 2026-05-26
Shrink and minify builds

formatBCE:
People complaining on APK grow. Minified, split by arch.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -45,7 +45,8 @@ android {
     buildTypes {
         release {
             signingConfig = signingConfigs.getByName("release")
-            isMinifyEnabled = false
+            isMinifyEnabled = true
+            isShrinkResources = true
             proguardFiles(
                 getDefaultProguardFile("proguard-android-optimize.txt"),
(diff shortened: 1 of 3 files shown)


androidApp/build.gradle.kts
```diff
@@ -60,10 +61,28 @@ android {
...
+                getDefaultProguardFile("proguard-android-optimize.txt"),
+                "proguard-rules.pro"
+            )
             signingConfig = signingConfigs.getByName("selfSigned")
         }
     }
+
+    // ABI splits for the GitHub-distributed APK only. The Play AAB path
+    // (bundleRelease) handles per-device delivery natively, so leave it alone.
+    splits {
```
seadowg, line 76: I'm not sure that we really need to bother with this complexity as (as you point out) the majority case will be people getting this from the Play Store and getting smaller downloads via the app bundle mechanism. I'd vote we either only distribute the universal APK or the `arm64-v8a` (until someone asks for the `armeabi-v7a`) to keep things simple for us and the users.

formatBCE: Yeah we can simplify.

PR #470, opened 2026-05-27
refactor: clean up Kotlin warnings and adjacent smells

teancom:
I finally had the time / bandwidth to dig through and figure out what all of the warnings during compilation were. Was able to fix most of them. Some of this isn't technically 'just' fixing compilation warnings, but came as a result of looking at the code that was generating the warning(s) and figuring out different approaches (that also happen to compile cleanly). Basically, trying to leave the codebase better than I found it. There shouldn't be any actual functional changes here, just cleanups.

Commit message:

Warnings: 33 → 2 (the remaining two are unfixable upstream — Compose Multiplatform's ObjC export of PathSegment.Type and the smallBinary debug-build info message). Detekt and Xcode build are clean.

Build / opt-ins / API renames
- composeApp build.gradle.kts: add '-Xexpect-actual-classes' to silence the 8 KT-61573 Beta warnings globally.
- KmpHelper.ios.kt: Ktor readBytes() → readRawBytes() (renamed API).
- SignalingMessage.kt: @OptIn(ExperimentalSerializationApi) on the list descriptor.
- MainViewController.kt, LogSharer.ios.kt: add BetaInteropApi to the file-level @OptIn (covers NSString.create cinterop calls).
- LogSharer.ios.kt: drop now-redundant @file:Suppress("EXPECT_ACTUAL_CLASSIFIERS_…") (compiler flag covers it); fix impossible 'as? String' cast — go through NSString.toString().
- KtorServiceClient.kt: drop unnecessary '?.' on non-null TransportState.Failed.error.
- SendspinWsHandler.kt, ItemListScreen.kt: remove redundant 'else' branches from two exhaustive 'when's.

Architectural cleanups
- AuthenticationViewModel: hoist Logger.withTag('AuthVM') into a private companion object (one Logger per class, not per VM instance); collapse duplicate init block (StateFlow's initial replay covers it); extract handleSessionState/onConnected/onDisconnected helpers. The init goes from ~60 lines of nested branches to a single collect(::handleSessionState).
- SendspinWsHandler: the inline Logger.withTag('WebSocketHandler') on network errors was creating a per-call Logger AND using a wrong tag AND dropping the throwable. Consolidated onto the file-level logger and pass the exception.
- WindowClass: DRY the three isAtLeast* Composables behind a private isAtLeastBreakpoint(widthDp) helper; migrate to currentWindowAdaptiveInfoV2().
- DataState.kt: new 'val <T> DataState<T>.dataOrNull: T?' extension that returns the payload for Data/Stale, null for Loading/Error/NoData — eliminates the duplicated 'is Data → data; is Stale → data' pattern scattered across the UI. Apply it in ItemListScreen (List type → '.orEmpty()') and CollapsibleQueue ('?: return@run' replaces the inner disambiguating whens).
- FlexibleUrlListSerializer (SignalingMessage.kt): use 'contentOrNull'
  + 'mapNotNull { (it as? JsonPrimitive)?.contentOrNull }' so a null or non-string element in the signaling payload no longer hard-throws.
- LogSharer.shareCrashLog: invert the file-exists guard to early return for shallower indentation; format the NSString.create call across lines for readability.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/auth/AuthenticationViewModel.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/auth/AuthenticationViewModel.kt
@@ -33,72 +33,46 @@ class AuthenticationViewModel(
     val password = MutableStateFlow("")
 
     init {
-        // Trigger initial load if already connected and awaiting auth
+        // sessionState is a StateFlow → emits its current value to the new collector
+        // immediately, so this single launch covers both initial-state handling and
+        // subsequent transitions.
         viewModelScope.launch {
-            val currentState = sessionState.value
-            if (currentState is SessionState.Connected) {
-                val dataConnectionState = currentState.dataConnectionState
-                if (dataConnectionState is DataConnectionState.AwaitingAuth) {
-                    // Only load providers if we're not in a failed state
-                    when (dataConnectionState.authProcessState) {
-                        is AuthProcessState.Failed -> {
-                            // Don't reload providers when auth failed
-                        }
-
-                        else -> {
-                            loadProviders()
-                        }
-                    }
-                }
-            }
-        }
-
-        // Auto-fetch providers when connected and awaiting auth
-        viewModelScope.launch {
-            sessionState.collect { state ->
-                Logger.d("AuthVM") { "SessionState changed: ${state::class.simpleName}" }
-                when (state) {
-                    is SessionState.Connected -> {
-                        val dataConnectionState = state.dataConnectionState
-                        Logger.d("AuthVM") { "DataConnectionState: ${dataConnectionState::class.simpleName}" }
-                        if (dataConnectionState is DataConnectionState.AwaitingAuth) {
-                            Logger.d("AuthVM") { "AwaitingAuth - checking auth process state" }
-                            // Only load providers if we're not in a failed state
-                            // (to avoid overriding error messages)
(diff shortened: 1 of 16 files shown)


teancom:
I'm double-checking some stuff, I want to make sure I'm *not* making any functional changes. Will flip it back once I'm sure.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/auth/AuthenticationViewModel.kt
formatBCE, line 1: I'm a bit worried about this class. Other things are okay, IMO.

teancom: I ended up reverting all of my changes to that file. I think it deserves a refactor but it would expand this PR a bit too much. I'll do a follow-up PR.

teancom:
Added two more commits and reverted part of the first one. Found some more json primitives that weren't being properly handled for nulls, and discovered that one of the fixes broke the pure common compilation tests, even though both Android and iOS compiled, so I fixed the fix. And then reverted all the changes to AuthenticationViewModel, I'll have a separate PR for that later. This one was getting large enough.

formatBCE (approved):

formatBCE:
Fuck, I may have created same `dataOrNull` helper for my changes to queue...
I just used class field, not extension.

PR #487, opened 2026-05-28
refactor(auth): rewrite AuthenticationViewModel as a flatMapLatest pipeline

teancom:
Separated out from #470 because it was going to be a bigger effort, this finishes the effort to get us down to only two compile-time warnings, while also refactoring and cleaning up AuthenticationManager. Tested on my Android phone, it seems to work great. If this refactoring *isn't* what you meant when you said "I'm a bit worried about this class" @formatBCE then I apologize. 😄

Commit message:

Replaces the procedural loadProvidersJob/loadingForWebRTC state machine with a reactive providers flow over sessionState + a retry trigger. flatMapLatest cancels any in-flight load on state change, closing the disconnect-vs-stale-result race the old code had (reviewer flagged this class on #470).

Extracts AuthCoordinator from AuthenticationManager so the VM depends on a narrow interface, not the manager (whose init spawns an auto-login coroutine that can't be unit-tested).

Because flatMapLatest now cancels suspend calls routinely, hardens all five cancellable suspend entry points in AuthenticationManager (getProviders, loginWithCredentials, getOAuthUrl, handleOAuthCallback, authorizeWithSavedToken) to rethrow CancellationException ahead of their broad catch -- otherwise cancellation is swallowed into a spurious AuthState.Error or an absorbed failure Result. startOAuthFlow and logout are exempt: no suspension point, so no cancellation to catch.

Adds kotlinx-coroutines-test, Turbine, and multiplatform-settings-test. 18 tests cover the projection, the structural race fix, retry semantics, the error path, login/OAuth/logout, and the cancellation contract on the three reachable suspend entry points (each verified red without the fix).

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -42,6 +42,7 @@ easyqrscan = "0.7.1"
 androidx-compose-ui-test = "1.11.2"
 robolectric = "4.16.1"
 detekt = "1.23.8"
+turbine = "1.2.1"
 
 
 [libraries]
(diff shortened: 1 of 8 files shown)


gradle/libs.versions.toml
```diff
@@ -42,6 +42,7 @@ easyqrscan = "0.7.1"
 androidx-compose-ui-test = "1.11.2"
 robolectric = "4.16.1"
 detekt = "1.23.8"
+turbine = "1.2.1"
```
formatBCE, line 45: I was afraid of this day, but it was inevitable.

formatBCE (approved):
Great reactive flow. Finally someone got to clear this legacy mess!

PR #489, opened 2026-05-29
fix(ios,audio): defer session activation and wire pause/resume sink

### LINKED ISSUE #279: Bug / iOS - Opening app (cleared out of background) claims audio playback even nothing is playing (yet)
It happens in first open/when not in ram. If i op MA then start playing something in tidal app and go back to MA app, tidal app keeps playing. If I close MA app from background/ram, start playing tidal app and then open MA app (without it being in background first) it claims audio playback while nothing is actually playing on the local ma player.

This is a problem if you for example want to open app to control external player, but also have audio running on local device from other app (that you don't want to have paused/stopped).

Testflight 1.0 (10)

**How to reproduce**

1. Play music on Spotify 
2. Open fully closed ma app
3. Ma pauses Spotify 
4. Manually starting playback again 
5. Moving  ma into background 
6. Opening again won’t pause audio

### LINKED ISSUE #386: Issues when handig off audio output to other app
## Summary
When handing of the audio from MA app to another app, the app is not pausing stream, because of this the app is also unable to reclaim the audio when manually pressing pause and then play again, need to remove app from multitasking and re-open to reclaim audio.

Closely related to: https://github.com/music-assistant/mobile-app/issues/279

## Devices, app and server information
- iPhone 17 - iOS 26.5
- App build: TestFlight 1.0 (25)
- MA Server 2.8.7 Stable

## Not handing off audio correctly
**How to reproduce**
1. Start playing music on local MA player
2. Move app to background, start playback of audio in other app
3. Move back to MA app, you'll notice playback of the stream local MA player has not been paused when other app claimed the audio (first part of the problem)
4. When manually pausing the local stream, and then manually pressing play again, MA app cant reclaim the on device audio output (second part of the problem). App is only able to reclaim audio after app has been fully closed (out of ram).

**Expected behaviour**
1. Start playing music on local MA player
2. Move app to background, start playback of audio in other app. At this stage the local MA player should be paused.
3. After moving back to app, playback of other app should stay active.
4. When pressing play again on local stream, MA app should reclaim audio output.

teancom:
Hopefully this clears up a lot of issues @remon1496 was seeing. Tested pretty thoroughly on my local iPhone, it no longer steals audio from Spotify on launch until you actively hit 'play' on the local player, it can resume after being paused by Spotify (and other audio apps).  And specifically tested that you can control other devices without stopping other apps from playing back locally (i.e., keep listening to Spotify on your phone while controlling your living room streamer).

Commit message:

Activate the audio session on real playback instead of at launch, so opening the app no longer claims audio from other apps. Implement pauseSink/resumeSink/flush on iOS: pause tears the queue down with a shouldPlay gate that drops in-flight audio, so resume reclaims the session and rebuilds cleanly instead of replaying stale buffers.

Also some whitespace cleanup in the .swift files while I was there.

Closes #279
Closes #386

### DIFF
--- a/composeApp/src/iosMain/kotlin/io/music_assistant/client/player/MediaPlayerController.ios.kt
+++ b/composeApp/src/iosMain/kotlin/io/music_assistant/client/player/MediaPlayerController.ios.kt
@@ -52,9 +52,9 @@ actual class MediaPlayerController actual constructor(platformContext: PlatformC
         return data.size
     }
 
-    actual fun pauseSink() { /* no-op on iOS */ }
-    actual fun resumeSink() { /* no-op on iOS */ }
-    actual fun flush() { /* no-op on iOS */ }
+    actual fun pauseSink() { PlatformPlayerProvider.player?.pauseSink() }
+    actual fun resumeSink() { PlatformPlayerProvider.player?.resumeSink() }
+    actual fun flush() { PlatformPlayerProvider.player?.flush() }
 
     actual fun stopRawPcmStream() {
         PlatformPlayerProvider.player?.stopRawPcmStream()
--- a/composeApp/src/iosMain/kotlin/io/music_assistant/client/player/PlatformAudioPlayer.kt
+++ b/composeApp/src/iosMain/kotlin/io/music_assistant/client/player/PlatformAudioPlayer.kt
@@ -22,6 +22,16 @@ interface PlatformAudioPlayer {
      * usePinned bulk-copy, avoiding a byte-by-byte Swift interop loop.
      */
     fun writeRawPcmNSData(data: NSData)
+
+    /** Pause output (user pause / interruption / re-phase). */
+    fun pauseSink()
+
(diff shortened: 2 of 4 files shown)


formatBCE:
Walt, should we resume after focus was stolen by other audio app? I thought this behaviour is reserved for calls in auto...

teancom:
So it actually already worked as we'd want, in terms of *automatic* resumption. Starting playback in, say, Spotify would *not* cause MA to resume after Spotify stops / quits, because iOS doesn't hand MA an `.ended` if another audio app stops, it only does after things like timer's finishing, Siri interrupting, etc. But! There is the `.shouldResume` gate that we weren't paying attention to so it was relying on iOS never sending us an `.ended` while `.shouldResume` was false (I have no idea what circumstances that would happen under, I wasn't able to get it to do that) but still, in a belt-and-suspenders sort of way we would use it. Tested the various flows again, everything is still working properly, at least as far as I've been able to tell.

formatBCE (approved):

PR #490, opened 2026-05-29
Improve the structure of the app documentation and include documentation on connecting via a proxy.

### LINKED ISSUE #174: Docs - Connection to server - Proxied
Create doc to explain the proxied connection and login.
Provide Nginx instructions.
Provide screenshots.

remon1496:
Closes #174


seadowg (changes requested):
I'm not sure if this is something we should document at this level for the app as it's really for Home Assistant itself. Additionally, there are many different ways to connect to a service in your local network remotely (this is really dynamic DNS with an nginx proxy running in HAOS rather than **just** a proxy for example) and I wouldn't really want to advocate for any of them over others given they are all littered with their own (kind of nightmarish) security implications.

Maybe we could just add a small section to the existing connection docs with an example of connecting to MA remotely as well as local connections? Something like:

> | **Server host** | The hostname or IP address of your Music Assistant server e.g. `homeassistant.local`<sup>*</sup>, `192.168.1.2` or a remote address like `ma-app.duckdns.org`. |
| **Port** | The port your Music Assistant server is listening on (default: `8095`). |
| **Use TLS (wss://)** | Enable this if your server uses a secure (TLS) connection. |

And then we could just mention that there are solutions for connecting remotely like Home Assistant Cloud and Duck DNS.

remon1496:
@seadowg I agree there are many different (none of them ideal) ways. Thats why intro states:

> The guide below covers one way to achieve this using the HA Duck DNS integration and the Nginx Proxy Manager app, but any proxy method will work.

The reason I chose to work on this doc first was the thought process of WebRTC performance (in-app) not being that great and at least providing a guide (specific to MA) for remote access without using WebRTC.

But I'm happy to move all connection methods (no in-depth details) into a single doc to simplify it, and we don't advocate one remote access option over others :)

formatBCE (approved):

formatBCE:
@seadowg how it looks now to you?

seadowg:
Love it! Thanks for listening to my ranting.

seadowg (approved):

PR #493, opened 2026-05-29
Fixed library stack. Changed the lib items layout.

formatBCE:


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/MultiBackStack.kt
```diff
@@ -36,14 +36,15 @@ class MultiBackStack<T : NavKey>(private val backStacks: List<MutableList<T>>) {
 
     /**
      * Switch to [index] back stack, optionally replacing its contents with a single
-     * [replacementBottom] entry. Used to redirect navigation into another tab's stack
+     * [pushOnTop] entry. Used to redirect navigation into another tab's stack
      * (e.g. tapping "All Albums" on Home jumps to the Library tab rooted at Albums).
      */
-    fun switchTo(index: Int, replacementBottom: T? = null) {
+    fun switchTo(index: Int, pushOnTop: T? = null) {
```
seadowg: I'm wondering if we should just open these "All <media type" links in the "Home" stack? That would simplify the code, and I think it feels like it'd make sense now that it's a separate screen (as opposed to the combine tab "hub" library screen before).

formatBCE: > I'm wondering if we should just open these "All <media type" links in the "Home" stack? That would simplify the code, and I think it feels like it'd make sense now that it's a separate screen (as opposed to the combine tab "hub" library screen before).

Great suggestion. We basically unbind stacks with it. Can remove that switchTo method.

androidApp/src/test/kotlin/io/music_assistant/client/support/pages/LibraryPage.kt
```diff
@@ -61,7 +63,9 @@ class LibraryPage(composeTestRule: ComposeTestRule) :
     }
 
     fun clickGenres(): ItemListPage {
-        return clickType(Res.string.media_type_genres.get())
+        val type = Res.string.media_type_genres.get()
+        composeTestRule.onNodeWithText(type).onParent().performScrollToIndex(7)
```
seadowg, line 67: There's an `inScrollable` helper in `Scrollable.kt` that can help with this in the future!

formatBCE: I feel like learning tests you wrote is another framework to know. :)

PR #508, opened 2026-06-01
fix(queue): load items when queue metadata lands, not on players-only emission

teancom:
This came up as I was testing some unrelated queue stuff. I would intermittently get a blank / empty queue on fresh launch even when I know the server had items in the queue for my device. As soon as I added something to the queue, everything else would appear. Turned out to be a race from the debounce that I was occasionally hitting.

Commit message:

The startup item fan-out gated on _playersData being Data, but the _queueInfos to _playersData combine is debounced: when players arrive more than EVENT_DEBOUNCE before queues, the first Data emission carries null queueInfo for every player, so refreshAllPlayersQueueItems skips them all and never retries. The queue then shows empty until something forces a refetch.

Gate on queueInfo actually being merged in, and cancel-and-replace the gate job so a never-resolving wait (zero-queue session) can't accumulate one suspended _playersData collector per reconnect.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
@@ -227,7 +227,6 @@ class MainDataSource(
 
     private var watchJob: Job? = null
     private var updateJob: Job? = null
-    private var queueItemsGateJob: Job? = null
 
     init {
         // Mirror optimistic-bump stamps into `_queueInfos` so the gate sees them.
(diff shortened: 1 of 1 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -227,6 +227,7 @@ class MainDataSource(
 
     private var watchJob: Job? = null
     private var updateJob: Job? = null
+    private var queueItemsGateJob: Job? = null
```
formatBCE: This field is not needed, since it is only assigned and canceled in `init`, so actually it never will effectively be canceled - thus rendering the field useless. Just modify `first` lambda to  add `&& state.data.any { it.queueInfo != null }`

formatBCE (changes requested):

formatBCE (approved):

PR #517, opened 2026-06-01
Fix back gesture for expanded player

### LINKED ISSUE #514: back button/gesture broken for expanded player
## Context
 - Operating system and version: Android
 - App version: debug-2026.05.28.0336
 - Music Assistant server version: 2.8.8

### Problem description

Using the back gesture on the expanded player doesn't close it: it pops the stack "underneath". There is an exception if the user is on the Home screen.

### Steps to reproduce

1. Open the app
2. Click on a media item like an album
3. Expand the player
4. Use the back gesture/button

The player doesn't close. Closing it via the button reveals you have returned to the home screen.

### Expected behavior

The player should be closed when the back gesture is used.

seadowg:
Closes #514

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/CustomNavDisplay.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/CustomNavDisplay.kt
@@ -1,0 +1,92 @@
+package io.music_assistant.client.ui.compose.nav
+
+import androidx.compose.animation.AnimatedContentTransitionScope
+import androidx.compose.animation.ContentTransform
+import androidx.compose.animation.SharedTransitionScope
+import androidx.compose.animation.SizeTransform
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.navigation3.runtime.NavEntry
+import androidx.navigation3.scene.Scene
+import androidx.navigation3.scene.SceneDecoratorStrategy
+import androidx.navigation3.scene.SceneInfo
+import androidx.navigation3.scene.SceneStrategy
+import androidx.navigation3.scene.SinglePaneSceneStrategy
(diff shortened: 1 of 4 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/CustomNavDisplay.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.navigation3.ui.defaultTransitionSpec
+import androidx.navigationevent.NavigationEvent
+import androidx.navigationevent.compose.NavigationBackHandler
+import androidx.navigationevent.compose.rememberNavigationEventState
+
+/**
+ * Custom version of [NavDisplay] that accepts (and handles the boilerplate for) [entries], but
+ * also allows the back handling to be enabled and disabled.
+ *
+ * An issue has been filed about this option being currently missing from Navigation 3: https://issuetracker.google.com/u/2/issues/518651940
```
seadowg, line 29: There's potentially a way to deal with this use case using the "scenes" concept, but it feels like a lot to dive into to just be allowed to display something over the `NavDisplay` and have it sometimes handle back instead.

formatBCE: Damn, this gets complicated...

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/CustomNavDisplay.kt
seadowg: Yeah. Like with Navigation 1 and 2, I think Navigation 3 is also running into just how hard it is to generalize "navigation" as a concept. Hopefully they'll have at least a better workaround suggestion as a reply to my issue.

PR #529, opened 2026-06-02
Add "Play album from here" option

seadowg:
Work towards #535

This just adds the long press option for the moment (no default click setting), but it also adds some changes like the `PlayHandler` type alias that it'd nice to be get in to avoid conflicts.


formatBCE (commented):
So now it's just default click behaviour?

What if I want to play only exact track?

Why not adding "Play from here" as we were discussing?

seadowg:
@formatBCE so with the new settings for default action added, I'll update this to be an option in there rather than the default. Sound good?

formatBCE:
> @formatBCE so with the new settings for default action added, I'll update this to be an option in there rather than the default. Sound good?

Yes, that will be awesome

formatBCE:
@seadowg ready to merge it?

seadowg:
@formatBCE sorry hadn't got the action itself in: needed to solve some problems that came up from getting the branch back up to date.

composeApp/src/commonMain/composeResources/values/strings.xml
```diff
@@ -15,15 +14,26 @@
   <string name="action_mark_unplayed">Mark as unplayed</string>
   <string name="action_ok">OK</string>
   <string name="action_pause">Pause</string>
+  <string name="action_customize">Customize…</string>
   <string name="action_play">Play</string>
   <string name="action_play_all">Play all</string>
   <string name="action_play_next">Play next</string>
   <string name="action_play_now">Play now</string>
+  <string name="action_play_album_from_here">Play album from here</string>
```
formatBCE, line 22: What about playlist?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/ItemAction.kt
```diff
@@ -38,7 +40,7 @@ import org.jetbrains.compose.resources.StringResource
 sealed class ItemAction(val kind: Kind) {
     enum class Kind { PLAYBACK, OTHER }
 
-    data class Play(val queueOption: QueueOption) : ItemAction(Kind.PLAYBACK)
+    data class Play(val queueOption: QueueOption, val fromHereInAlbum: Boolean = false) : ItemAction(Kind.PLAYBACK)
```
formatBCE: I'd maybe rather create another synthetic queue option - this looks very unnecessary...

composeApp/src/commonMain/composeResources/values/strings.xml
seadowg: Just album in this PR! Will follow up with playlist and default click settings.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/ItemAction.kt
seadowg: Yeah, I agree actually. Will do that.

formatBCE (approved):
Great, huge job

remon1496:
'Play album from here' string is hardcoded. Could you extact it in PR for playlists maybe? Nevermind, it's merged. Will open issue on it.

https://github.com/music-assistant/mobile-app/issues/577

PR #546, opened 2026-06-06
Add Audiobook chapter UI to the queue

teancom:
I've been working on this for a few days, and it's still just a v1. But the tl;dr is: I added 'chapter' entries for audiobooks in the queue view, so people can skip directly to a chapter. It also 'highlights' the current chapter during playback. And it only shows for the *current* audiobook, so your queue isn't gunked up if you have multiple books in it. I still need to add the 30s +/- and other audiobook-specific UI, that isn't this branch. Tested on my iPhone (extensively) and my Samsung (a bit).

To see it in action, using something like the AudioBookshelf provider, start playback on an audiobook with chapters, then open the queue.

Commit message:

Adds tap-to-seek chapter rows under the current audiobook (QueueDisplayRows flattens queue items + chapters; PlayersPager and CollapsibleQueue wire the live position). Rows are keyed by list index rather than server-supplied chapter position, which isn't guaranteed unique, and the active chapter is highlighted via derivedStateOf so the queue body doesn't recompose on every ~500ms position tick.

The one-shot open-scroll now waits for the first resolved live position before targeting a chapter, so opening the queue just after a chapter boundary lands on the chapter playing now instead of the previous one; chapter-less queues still scroll to the current item immediately. Unit tests cover row building, active-chapter selection, chapter-seek rounding, and duration formatting.


formatBCE:
Good stuff!
Can you maybe remove that chapters button that I added to the audiobook row in queue for sudo-chapters access? It's not needed anymore, and will return back the uniformity. :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/CollapsibleQueue.kt
```diff
@@ -506,6 +610,65 @@ fun Queue(
...
+            )
+        }
+    }
+}
+
+/**
+ * Human chapter length, or null below one second (nothing worth showing).
+ * Seconds under a minute, whole minutes under an hour, else `Hh Mm` / `Hh`.
+ */
+internal fun formatChapterDuration(seconds: Double): String? = when {
```
formatBCE: Can't we use `Double?.formatDuration(unit: DurationUnit)` from Ext.kt? Or at least move this method there? (I'd rather just reuse that one).

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/QueueDisplayRows.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+/**
+ * Flattens queue items into display rows, inserting chapter rows only under the
+ * current item when it is an audiobook with chapters. Position-independent: does
+ * not encode active state, so the result is stable across playback ticks and
+ * safe to `remember` keyed only on [items] and [currentItemId].
+ */
+internal fun buildDisplayRows(
```
formatBCE: I love using extension functions if they make sense:
```
internal fun List<QueueTrack>.buildDisplayRows(currentItemId: String?) =
    this.flatMapIndexed { queueIndex, item ->
        buildList {
            add(QueueDisplayRow.QueueItem(queueIndex, item))
            this.takeIf { item.id == currentItemId }?.let {
                (item.track as? Audiobook)?.chapters.orEmpty()
                    .forEachIndexed { chapterIndex, chapter ->
                        add(
                            QueueDisplayRow.ChapterItem(
                                parentQueueIndex = queueIndex,
                                parentQueueItemId = item.id,
                                chapterIndex = chapterIndex,
                                chapter = chapter,
                            )
                        )
                    }
            }
        }
    }
```

And use it like `queueItems.buildDisplayRows(currentId)`

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/QueueDisplayRows.kt
```diff
@@ -0,0 +1,69 @@
...
+                parentQueueIndex = queueIndex,
+                parentQueueItemId = item.id,
+                chapterIndex = chapterIndex,
+                chapter = chapter,
+            )
+        }
+    }
+
+/** Last chapter whose start is at or before [position]; null if [position] precedes all. */
+internal fun activeChapter(chapters: List<Chapter>, position: Double): Chapter? =
```
formatBCE: same, you can use `List<Chapter>.activeChapter(position: Double)`.

formatBCE (changes requested):

teancom:
Done! Great finds, thank you. And I wasn't sure about the bookmark icon before (wasn't 100% on why it was there, tbh) so yeah, pulled it and got to pull a good chunk of code along with it.

formatBCE (approved):

PR #551, opened 2026-06-06
feat(carplay): localize CarPlay strings via shared catalog (#542)

### LINKED ISSUE #542: Extract Strings for Apple CarPlay
## Problem

Apple CarPlay is currently shown in English, even though app is set to Dutch.

## Solution

Extract Apple CarPlay strings to they are translatable. Possibly also Android Auto, but didn't test that, maybe strings are already there.

teancom:
CarPlay UI was hardcoded in English regardless of app language. Route all CarPlay strings through the shared Compose string catalog (the same values/strings.xml Lokalise manages) so CarPlay follows the same translations as the rest of the app, with no second translation pipeline.

A new CarPlayStrings resolver (iosMain) reads strings for the active locale via getString; KmpHelper exposes it to Swift, which defers its first template build until the strings load because CarPlay template titles are immutable once constructed. Reuses existing media_type_*/nav_library/library_empty keys and adds 9 neutral, interface-agnostic keys so the upcoming Android Auto work can reuse them where applicable.

Closes https://github.com/music-assistant/mobile-app/issues/542

### DIFF
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -1,8 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources>
+  <string name="action_add_all_to_queue">Add all to queue</string>
   <string name="action_add_to_library">Add to library</string>
   <string name="action_add_to_playlist">Add to playlist</string>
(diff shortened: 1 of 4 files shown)


formatBCE (approved):

formatBCE:
Nice one!

PR #559, opened 2026-06-07
fix(ios,audio): preserve queued resume across backgrounded reconnect

### LINKED ISSUE #554: Still issues in 1.0 (33) that music stops playing when CarPlay announces navigation commands
## Context
 - Operating system and version: Home Assistant OS 2026.6.1
 - App version: 1.0 (33)
 - Music Assistant server version: Home-Asssitant addon, serverversion 2.8.9. Operating System 17.3

### Problem description

Still issues in 1.0 (33) that music stops playing when CarPlay announces navigation commands

### Steps to reproduce

The steps that we can take to reproduce your issue. Ideally format it as a list like:

1. Play music
2. Carplay navigation announces something
3. Music stops playing. I have to press play again to start the music

### Expected behavior

After the navigation announcement, the music should continue to play

[ma_client_logs.txt](https://github.com/user-attachments/files/28685996/ma_client_logs.txt)

teancom:
I *believe* this closes https://github.com/music-assistant/mobile-app/issues/554 as I'm no longer able to reproduce it. But oofdah, it can get tricky with all these state machines running around... A *little* more verbose than usual in the comments just to help future-me remember what each bit is supposed to be in charge of.

Commit 5915037 correctly re-issues a "play" when an audio interruption ends, but on CarPlay the interruption-pause stops playback, iOS drops the now-idle socket, and the transport reconnects into Disconnected.Backgrounded. That path ran stopSendspin() -> clearState(), which wiped the offline command queue -- discarding the very resume that drainCommandQueue() is meant to replay on reconnect. The teardown and the recovery were fighting, and the teardown won.

Split the two concerns clearState() had conflated: clearState() now only drops the optimistic local-player UI state, and a new clearCommandQueue() discards pending intent. stopSendspin() becomes a pure transport/pipeline teardown that preserves both, so a queued resume survives a transient background/reconnect and drains when the socket returns. Genuine resets (clearAllData, Sendspin disabled) clear them explicitly. This also keeps localPlayerData non-null across the teardown, closing a secondary race where a late .ended dropped the resume at onRemoteCommand's null check.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
@@ -665,10 +665,10 @@ private data class PlayerBuildInputs(
                         localPlayerRepository.onInitialPlayersReceived(hasLocalPlayer = false)
                     } else {
                         stopSendspin()
-                        // User turned Sendspin off — the local player is gone for good,
-                        // so drop its UI state and any queued offline commands.
+                        // User turned Sendspin off — the local player is gone for good.
+                        // stopSendspin() no longer resets it (transient teardowns must
+                        // preserve a queued resume), so clear it explicitly here.
                         localPlayerRepository.clearState()
-                        localPlayerRepository.clearCommandQueue()
                     }
                 }
             }
@@ -930,7 +930,6 @@ private data class PlayerBuildInputs(
         _queueInfos.update { emptyList() }
         positionTracker.clear()
         localPlayerRepository.clearState()
-        localPlayerRepository.clearCommandQueue()
         // Note: _providersIcons deliberately NOT cleared (static data)
     }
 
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -665,6 +665,10 @@ class MainDataSource(
                         localPlayerRepository.onInitialPlayersReceived(hasLocalPlayer = false)
                     } else {
                         stopSendspin()
+                        // User turned Sendspin off — the local player is gone for good,
+                        // so drop its UI state and any queued offline commands.
+                        localPlayerRepository.clearState()
```
formatBCE, line 671: It seems like you split the method to two, but you still call them together everywhere. Makes sense putting them back together, and then the only change is getting rid of that call on line 1139?

teancom: Sigh. Duh.

formatBCE (approved):

formatBCE:
I will keep my eye on it on Android side, but should be fine.

PR #561, opened 2026-06-07
Fix two row bar scroll

### LINKED ISSUE #547: Double topbar in Child Libraries causes erratic scroll speed when collapsing/expanding
## Context
 - Operating system and version: iOS 26.5.1
 - App version: Internal TestFlight 1.0 (32)
 - Music Assistant server version: 2.8.9 Stable

### Problem description
The Library Redesign introduced a second topbar row in Child Libraries. Both topbars collapse and expand simultaneously when scrolling, which doubles the effective scroll distance during this transition — making the scroll feel erratic and faster than expected.

### Steps to reproduce
1. Navigate to a Child Library.
2. Scroll down to collapse the topbars, or scroll back up to reveal them.
3. Observe both topbars animating simultaneously.

### Expected behavior
Scrolling speed should remain consistent throughout. Each topbar should respond to scroll input independently so their combined effect matches the normal scroll rate — for example, by having each topbar consume half the scroll delta.

https://github.com/user-attachments/assets/1c664655-106d-4082-afb9-596e9562aedb

formatBCE:
Fixes #547 and not just that.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/Screen.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/Screen.kt
@@ -1,5 +1,6 @@
 package io.music_assistant.client.ui.compose.nav
 
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Surface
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/Screen.kt
```diff
@@ -1,29 +1,57 @@
...
 @Composable
-fun Screen(topBar: @Composable (TopAppBarScrollBehavior) -> Unit, content: @Composable () -> Unit) {
+fun Screen(topBar: @Composable () -> Unit, content: @Composable () -> Unit) {
     Surface {
         val scrollBehavior = TopAppBarDefaults.enterAlwaysScrollBehavior()
         Column(
             modifier = Modifier.nestedScroll(scrollBehavior.nestedScrollConnection),
         ) {
-            topBar(scrollBehavior)
+            Box(modifier = Modifier.collapsingTopBar(scrollBehavior)) {
```
seadowg, line 34: This change does result in content scrolling directly under the status bar. Here's how it looks when it's collapsed:

Image

...and here it is before when the `TopAppBar` is in charge of the collapsing:

Image

Was that intentional?

formatBCE: It wasn't intentional. But I noticed it. Well, it can look weird, despite me being amused by it... :(

seadowg: Good to know! I can file an issue.

Issue #535, opened 2026-06-03
Add "Play album/playlist from here" actions to option menus


seadowg:
## Problem

There's no "Play album/playlist from here" option when long pressing a track on an album/playlist screen. This means I have to cancel the menu and then tap the track to have that happen, and it's less clear that that's possible behaviour.

## Solution

There should be an explicit action available. These should also both be available default click options.

formatBCE:
I argue about default click. It should be an option.
Also, if, say, we click track in track list, not in album - what would happen?
It's very confusing functionality.

seadowg:
@formatBCE I've updated the issue to make it clear these should be available as default click options as well as long press options as discussed.

formatBCE:
What's left here? Refactoring?

seadowg:
@formatBCE they're not present in the default click action settings yet.

PR #575, opened 2026-06-09
Add "Play playlist from here" action

seadowg:
Work towards #535

This just adds the action: next PR will add default click action settings and close the issue out.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/ItemAction.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/ItemAction.kt
@@ -3,7 +3,6 @@ package io.music_assistant.client.ui.compose.common.items
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.PlaylistAdd
 import androidx.compose.material.icons.filled.AddToQueue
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/ItemAction.kt
```diff
@@ -41,7 +43,7 @@ sealed class ItemAction(val kind: Kind) {
     enum class Kind { PLAYBACK, OTHER }
 
     data class Play(val queueOption: QueueOption) : ItemAction(Kind.PLAYBACK)
-    data object PlayFromHere : ItemAction(Kind.PLAYBACK)
+    data class PlayFromHere(val isPlaylist: Boolean) : ItemAction(Kind.PLAYBACK)
```
seadowg, line 46: I think this should use a type enum/ADT - this was just my first pass to get the test passing. However, I'd like to clean up where the boundary is between passing the "parent" (`AppMediaItem`) item around and using `ClickContext`/`SubItemClickContext` (and even `DefaultClickAction`) for these "from here" actions, and I think makes most sense to do as part of adding the default action settings next. Then, I can think about how all this fits together without holding up people using the action itself!

formatBCE (approved):

PR #599, opened 2026-06-12
Don't use server dynamic colors

seadowg:
As discussed, this removes the server colors (which creates inconsitencies) and just uses the locally calculated colors everywhere. I've also made sure that animations only run once for a navigation entry so you don't see colors pop back in when going back.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemHeader.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemHeader.kt
@@ -170,11 +170,9 @@ internal fun ItemTopBar(
     // tween on top of our color animation and visibly lag the header. Transparent container =
     // single-pass color change, in lockstep with the header gradient.
     Box(
-        modifier = Modifier.apply {
-            colors?.dominant?.inactive()?.let {
-                background(Brush.verticalGradient(listOf(it, MaterialTheme.colorScheme.surface)))
-            }
-        },
+        modifier = Modifier.background(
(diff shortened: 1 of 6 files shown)


seadowg:
~~@formatBCE a compromise here could be to keep the gradient in (in cases where can calculate one) as we never show that solid so the colors being "close, but not quite" doesn't really create as much visual weirdness. I'd just need to fix the reanimating when we return to screens, but I had a solution for that before I noticed the color clashes.~~

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemHeader.kt
```diff
@@ -90,11 +87,17 @@ fun ItemHeader(
     Column(
         modifier = Modifier
             .fillMaxWidth()
-            .background(
-                Brush.verticalGradient(
-                    listOf(colors.dominant.inactive(), MaterialTheme.colorScheme.surface),
-                ),
-            ),
+            .let {
```
formatBCE, line 90: I tend to use `apply` in cases like this - when we mutate object itself inside of lambda. Then you don't need `it` reference. :)

formatBCE: And then we use 
```
colors?.dominant?.inactive()?.let {
    background(Brush.verticalGradient(listOf(it, MaterialTheme.colorScheme.surface)))
 }
```

formatBCE (commented):
Good code. If we decide to remove that - let's merge.

seadowg:
Closing in favour of another PR

PR #600, opened 2026-06-12
Add "Play from here" as default click option

### LINKED ISSUE #535: Add "Play album/playlist from here" actions to option menus

seadowg:
Closes https://github.com/music-assistant/mobile-app/issues/535

This adds the album/playlist default option and also reworks how the action icon/title is determined so that it uses `ClickContext` instead of needing to pass the parent `Album`/`Playlist` around.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
@@ -27,19 +27,19 @@ val carBrowsableKinds: List<ItemKind> = listOf(
 )
 
 /** Today's hard-coded pair — the default bulk actions for a browsable kind with no stored config. */
-val defaultCarBulkActions: List<DefaultClickAction> = listOf(
-    DefaultClickAction.PLAY_NOW,
-    DefaultClickAction.ADD_TO_QUEUE,
+val defaultCarBulkActions: List<DefaultClickOption> = listOf(
+    DefaultClickOption.PLAY_NOW,
+    DefaultClickOption.ADD_TO_QUEUE,
 )
 
 /**
  * Whether this action can be offered/honored on [platform] for [kind]. Intersects the kind
- * applicability ([DefaultClickAction.appliesTo]) with per-platform capability. Both the Settings
+ * applicability ([DefaultClickOption.appliesTo]) with per-platform capability. Both the Settings
  * UI (offered options) and the car-side dispatch consult this, so an unsupported stored action is
  * silently skipped rather than mis-dispatched. The platform arms are identical today; the seam
  * exists so a future CarPlay/AA limitation lands in one place.
  */
-fun DefaultClickAction.isCarSupported(platform: CarPlatform, kind: ItemKind): Boolean {
+fun DefaultClickOption.isCarSupported(platform: CarPlatform, kind: ItemKind): Boolean {
     if (!appliesTo(kind)) return false
     return when (platform) {
         CarPlatform.ANDROID_AUTO -> true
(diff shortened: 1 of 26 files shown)


seadowg:
@formatBCE this is currently failing because adding a new `DefaultClickAction` means handling it in the Auto/Carplay parts of the app. I don't have a car with Android Auto (or any car for that matter), but I'm assuming I try out the UI that `AutoLibrary` deals with in an emulator?

formatBCE:
You can use Android DHU emulator for Auto.

seadowg:
@formatBCE do you have a set of steps for actually getting to the library UI displayed using `AutoLibrary`? I've never used the Auto interface in my life so it's all a bit strange. Maybe I'm misinterpreting things and it only gets used for voice?

formatBCE:
> @formatBCE do you have a set of steps for actually getting to the library UI displayed using `AutoLibrary`? I've never used the Auto interface in my life so it's all a bit strange. Maybe I'm misinterpreting things and it only gets used for voice?

Nah, not just voice! :) 
To test it without the car, you need to install Headless Unit (it's program that emulates Android Auto). Then you just launch it, connect your phone to computer via USB - and it shows you Auto interface emulated via Headless Unit (DHU).

Installation happening through Android SDK manager. Then you do something like
`/Users/{your_user}/Library/Android/sdk/extras/google/auto/desktop-head-unit --usb`

Also check https://github.com/music-assistant/mobile-app/blob/dev/docs/ANDROID-AUTO.md#1-enabling-android-auto-for-sideloaded-builds on how to work with debug builds.

composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
```diff
@@ -27,50 +27,56 @@ val carBrowsableKinds: List<ItemKind> = listOf(
...
  * exists so a future CarPlay/AA limitation lands in one place.
  */
-fun DefaultClickAction.isCarSupported(platform: CarPlatform, kind: ItemKind): Boolean {
+fun DefaultClickOption.isCarSupported(platform: CarPlatform, kind: ItemKind): Boolean {
     if (!appliesTo(kind)) return false
-    return when (platform) {
-        CarPlatform.ANDROID_AUTO -> true
-        CarPlatform.CARPLAY -> true
+
+    return if (this == DefaultClickOption.PLAY_FROM_HERE) {
```
seadowg, line 45: @formatBCE currently, I've just blocked `PLAY_FROM_HERE` in Auto/CarPlay so that this can be ready to merge without me needing to set up my Auto dev environment to check features there. That way I can file a follow-up issue to add support for cars that I (or someone with an existing car env) can do later.

Can also just leave this unmerged until I've got a bit more time as well if you prefer!

formatBCE: Yeah we can merge it. I fully appreciate your situation, no need installing DHU. :)
I will add these to Auto.

seadowg: Thanks! Yeah I'd love to dig and play around with it, but just trying to be realistic with how much time I can give.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsViewModel.kt
```diff
@@ -400,19 +400,20 @@ class ItemDetailsViewModel(
...
+            onPlayClick(it, option, radio, false)
         }
     }
 
     fun onPlayClick(
         item: AppMediaItem,
         option: QueueOption,
         radio: Boolean,
-        parent: AppMediaItem? = null,
+        formHereInParent: Boolean,
```
formatBCE: from* ?

seadowg: Urgh. My most common typo.

formatBCE (approved):
Nice with context!

PR #628, opened 2026-06-16
feat(logging): make in-app shared logs useful for field debugging

teancom:
This is a direct result of me trying to debug some CarPlay issues, which by definition mean my phone can't be connected to the Xcode debugger, so I have to rely on what's persisted to the system log + 'share logs' inside the app.

Commit message:

Make the in-app diagnostic buffer able to reconstruct network and playback issues for a longer period of time. Clean up logging overall.

- InMemoryLogWriter: own Info severity floor (independent of the console/verbose level) and larger capacity, retaining ~30+ min of Info+ history
- Event.kt: stop dumping full payloads for unmodeled server events (tasks_updated carries multi-KB embedded server logs); record the type at Debug instead.
- iOS: new NativeLog bridge routes Swift logs through Kermit (in-memory buffer + os.Logger) instead of Xcode-only print(); convert NativeAudioController, NowPlayingManager, and AudioDecoders with state/commands/errors at Info+ and per-packet/hot-path detail at Debug.
- Kotlin Info audit: promote commands sent/received, websocket close, stream metadata, and token-clear-on-auth-failure to Info; demote the 2s state-report heartbeat to Debug (with an explicit Info at the Buffering -> Synchronized transition); reclassify a stray Logger.e seek log.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/DirectTransport.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/DirectTransport.kt
@@ -106,7 +106,7 @@ class DirectTransport(
                 }
             } catch (@Suppress("SwallowedException") e: ClosedReceiveChannelException) {
                 // Expected on graceful close — the channel-closed exception type IS the signal.
-                logger.d { "WebSocket connection closed" }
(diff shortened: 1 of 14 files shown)


formatBCE (approved):
Really helpful refactoring.

PR #629, opened 2026-06-16
Reset screens when tapping nav item on root

### LINKED ISSUE #408: Reset screen when pressing nav item on root screen
## Problem

I'd like to be able to quickly return the nav stack I'm viewing to its original state. Right now if I'm deep in a stack in the Home/Library/Search, I can press the relevant nav item button to return to the root screen, but the screen itself will still be in the state it was before.

## Solution

Pressing the nav item while already on the root should "reset" that item: the view should be scrolled back to the top. On "Search", the results and input should be cleared as well.

seadowg:
Closes #408

One difference from the issue:

> On "Search", the results and input should be cleared as well.

After playing with just resetting scroll, I was less confident we wanted this. It's quite nice to be able to quickly reset back to the top of the screen and change filters or add to the search field. We can always add the additional feature (and complexity) of clearing the search field as well later if we still feel like we want it.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
@@ -288,14 +288,6 @@ private fun LandingPageTopBar(
     TopAppBar(
         title = { Text(stringResource(Res.string.nav_home)) },
         actions = {
-            if (!editMode) {
-                IconButton(onClick = onRefresh) {
-                    Icon(
-                        imageVector = Icons.Default.Refresh,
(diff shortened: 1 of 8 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
```diff
@@ -289,6 +296,15 @@ private fun LandingPageTopBar(
                     ),
                 )
             }
+
+            if (!editMode) {
```
seadowg, line 300: I put these in the wrong order, and have kept meaning to flip them back

formatBCE (approved):
Nice work!

PR #634, opened 2026-06-17
Simplify Android build setup

seadowg:
This removes the split APK build and just build a universal one. I don't think the download size difference isn't really worth the complexity here for a Github APK release - the actual Play Store App Bundle will always be much smaller. Alternatively, we could just build an `arm64-v8a` APK, but that's still extra configuration.

I've also removed some unused config/workflow steps.

### DIFF
--- a/.github/workflows/debug-apk-release.yml
+++ b/.github/workflows/debug-apk-release.yml
@@ -48,7 +48,7 @@ jobs:
           keyPassword=${{ secrets.KEY_PASSWORD }}
           EOF
 
-      - name: Build self-signed release APKs (per-ABI + universal)
+      - name: Build self-signed release APK
         run: ./gradlew :androidApp:assembleSelfSignedRelease
 
       - name: Get vars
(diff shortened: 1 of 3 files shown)


.github/workflows/debug-apk-release.yml
```diff
@@ -38,17 +38,7 @@ jobs:
           mkdir -p ~/.android
           echo "${{ secrets.DEBUG_KEYSTORE_BASE64 }}" | base64 -d > ~/.android/debug.keystore
 
-      - name: Restore release keystore
```
seadowg, line 41: We might actually want to use these for the Github release, but I've started a separate conversation about that.

.github/workflows/debug-apk-release.yml
```diff
@@ -80,17 +70,9 @@ jobs:
...
 
-            ### Installation
-            Pick the APK that matches your device:
-            - **`*-arm64-v8a-*.apk`** — recommended for all modern Android devices (2017+).
-            - **`*-armeabi-v7a-*.apk`** — older 32-bit ARM devices only.
-            - **`*-universal-*.apk`** — fallback containing all ABIs (larger).
-
             Signed with the consistent debug keystore, so installing over a previous build does not require uninstall.
           files: |
-            androidApp/build/outputs/apk/selfSignedRelease/androidApp-arm64-v8a-selfSignedRelease.apk
```
formatBCE, line 91: I did this to reduce APK size - users were complaining. It costs nothing, why remove? :)

seadowg: Ah OK I hadn't seen any complaints, but fair enough! My goal was just to have one APK to make it simpler for people downloading (and to simplify the config). How about we just do the `arm64-v8a` as a compromise (as I suggest in the description)?

formatBCE: Yeah v7 is waaaay outdated by now. :)

formatBCE (approved):

Issue #638, opened 2026-06-17
Reload Playlist Library after Playlist deletion


remon1496:
## Problem (closely related to #637)

When removing a playlist from library via the in app long-press 'remove playlist` menu option, the user doesn't really have a good idea if the playlist has been removed or not, it is.

At this point we still show the playlist library screen from the moment we initially loaded it (where the now deleted playlist still was present).

## Solution

I think one way to solve this is to request the playlist items again (like what happens when doing a page refresh). This will show the updated playlist library and now 'hide' the deleted playlist.

formatBCE:
fixed in #646

PR #646, opened 2026-06-17
Fixed new playlist and delete playlist flows.

### LINKED ISSUE #637: Close 'Create Playlist' dialog and refesh playlists library screen after playlist creation
## Problem (closely related to #638)

When creating a playlist from within the app via the in app `+ add playlist` option, We're presenting user with the `New playlist` dialog. In here user can enter the `name of the playlist` and then click `create`. After clicking the `create` button, the dialog is not dismissed, so the user doesn't really have a good idea if the playlist has been created or not, it is.

The playlist has now been created on the general `Music Assistant` provider (the built in MA library playlist manager works independently from other music providers and contain items from every 'third party' music provider). However, since we made an API call to server to create this playlist (and the playlist was actually created), we currently don't show the newly created playlist in the playlist library, because the playlist didn't excist yet on the moment we initialy loaded the playlists. I think one way to solve this is to request the playlist items again (like what happens when doing a page refresh). This will show the newly created playlist.

## Solution

After opening playlist dialog and, filling in playlist name and pressing the `create` button, close the `create playlist` dialog and re-load playlists from server so the newly created playlist is shown in app.

formatBCE:
Fixes #637

### DIFF
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -242,6 +242,7 @@
   <string name="sort_year">Year</string>
   <string name="toast_added_to_playlist">Added to %1$s</string>
   <string name="toast_error_add_playlist">Error adding to playlist</string>
+  <string name="toast_error_create_playlist">Error creating playlist</string>
   <string name="toast_error_mark_played">Error marking as played</string>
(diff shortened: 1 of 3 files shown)


remon1496:
I think this also fixed https://github.com/music-assistant/mobile-app/issues/638 ?

PR #651, opened 2026-06-18
Fix startup crash on Android < 12 (NoClassDefFoundError: AudioManager.OnModeChangedListener)

starbrightlab:
## Problem

The app crashes on launch on every device running **Android < 12 (API < 31)**, despite declaring `minSdk 28`:

```
java.lang.NoClassDefFoundError: Failed resolution of: Landroid/media/AudioManager$OnModeChangedListener;
  ... Caused by: java.lang.ClassNotFoundException: android.media.AudioManager$OnModeChangedListener
  at io.music_assistant.client.MainActivity.getDataSource
  at io.music_assistant.client.MainActivity.onCreate
```

`AudioManager.OnModeChangedListener` was added in **API 31**.

## Root cause

In `MediaPlayerController.android.kt`, `modeChangedListener` is an eager `val`:

```kotlin
@RequiresApi(Build.VERSION_CODES.S)
private val modeChangedListener = AudioManager.OnModeChangedListener { mode -> ... }
```

Being a non-lazy property, it's initialized in the constructor on **every** API level, and instantiating the listener resolves the API-31 type → `NoClassDefFoundError` on Android < 12. `@RequiresApi` only affects lint, not runtime. The `register`/`unregister` call sites are already correctly guarded with `SDK_INT >= S`, so the field initializer was the only unguarded reference.

## Fix

Make the listener `by lazy`, so the API-31 type is only referenced on first access — which happens exclusively inside `registerModeChangedListener()` (guarded by `SDK_INT >= S`). No behavioural change on API 31+.

```kotlin
@get:RequiresApi(Build.VERSION_CODES.S)
private val modeChangedListener by lazy { AudioManager.OnModeChangedListener { mode -> ... } }
```

## Testing

- Built `:androidApp:assembleDebug` and installed on a **Meta Portal (Android 10 / API 29)**.
- Before: instant crash on launch with the trace above.
- After: app launches normally and reaches the connection screen.
- No change on API 31+ (listener still registered/unregistered behind the existing SDK guards).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/composeApp/src/androidMain/kotlin/io/music_assistant/client/player/MediaPlayerController.android.kt
+++ b/composeApp/src/androidMain/kotlin/io/music_assistant/client/player/MediaPlayerController.android.kt
@@ -70,19 +70,24 @@ actual class MediaPlayerController actual constructor(platformContext: PlatformC
     // deliver AUDIOFOCUS_LOSS_TRANSIENT cleanly on incoming calls. OnModeChangedListener
     // (API 31+) fires on MODE_IN_CALL / MODE_IN_COMMUNICATION transitions and needs no
     // runtime permission, so we use it as a redundant signal.
-    @RequiresApi(Build.VERSION_CODES.S)
-    private val modeChangedListener = AudioManager.OnModeChangedListener { mode ->
-        val inCall = mode == AudioManager.MODE_IN_CALL ||
-            mode == AudioManager.MODE_IN_COMMUNICATION
-        if (inCall && shouldPlayAudio) {
-            logger.i { "Telephony mode=$mode — pausing server playback (focus backup)" }
-            shouldPlayAudio = false
-            pausedByFocusLoss = true
-            audioTrack?.pause()
-            onRemoteCommand?.invoke("pause")
+    // Lazy so the API 31 OnModeChangedListener type is only referenced on API >= 31.
+    // An eager initializer runs in the constructor on every API level and throws
+    // NoClassDefFoundError on Android < 12, even though registration is SDK-guarded.
+    @get:RequiresApi(Build.VERSION_CODES.S)
+    private val modeChangedListener by lazy {
+        AudioManager.OnModeChangedListener { mode ->
+            val inCall = mode == AudioManager.MODE_IN_CALL ||
+                mode == AudioManager.MODE_IN_COMMUNICATION
+            if (inCall && shouldPlayAudio) {
+                logger.i { "Telephony mode=$mode — pausing server playback (focus backup)" }
+                shouldPlayAudio = false
+                pausedByFocusLoss = true
+                audioTrack?.pause()
+                onRemoteCommand?.invoke("pause")
+            }
+            // Resume is driven by AUDIOFOCUS_GAIN only — mode can return to NORMAL before
+            // focus is restored, and we only auto-resume what we paused.
         }
-        // Resume is driven by AUDIOFOCUS_GAIN only — mode can return to NORMAL before
-        // focus is restored, and we only auto-resume what we paused.
     }
     private var isModeChangedListenerRegistered = false
 


seadowg:
@starbrightlab what are the reproduction steps for the crash? I just launched the app, logged in and played music on an API 30 emulator and didn't see a crash.

seadowg (changes requested):
https://github.com/music-assistant/mobile-app/pull/651#issuecomment-4740272534

starbrightlab:
! It reproduces on Android 10 (API 29). I hit it on a physical Meta Portal Go, instantly on launch, before the login screen renders:

 ```
 java.lang.NoClassDefFoundError: Failed resolution of: Landroid/media/AudioManager$OnModeChangedListener;
    at io.music_assistant.client.MainActivity.getDataSource
    at io.music_assistant.client.MainActivity.onCreate
```
**Caused by**: java.lang.ClassNotFoundException: android.media.AudioManager$OnModeChangedListener
**Repro**: install on an API 29 (Android 10) device/emulator (the app declares minSdk 28, so it's in range) and launch. It crashes on startup before login.

OnModeChangedListener is API 31, but MediaPlayerController creates it as an eager val in its constructor, so the type is resolved on every API level even though registerModeChangedListener() is correctly guarded behind SDK_INT >= S. I suspect your API 30 emulator didn't hit it because of a difference in ART's class-verification strictness vs API 29.

The fix just makes the field by lazy, so the API-31 type is only touched on first access (inside the already-guarded register path). There's no behavioral change on API 31+.

**Review by formatBCE — APPROVED** [review]
I think we can merge it. We will have to cut out Android 10 at some point - it's on End of Life stage, no point to support it if it's hard. But this one is minimal change, that won't really affect other versions - so it's okay. What do you think @seadowg ?

**Comment by seadowg** [comment]
@formatBCE I'm not able to reproduce the crash on API 29 or 30, so not sure if this fix is correct or not. Are you?

**Comment by starbrightlab** [comment]
The registration sites (lines 508, 521) are already correctly guarded with SDK_INT < S checks. You just missed that the field itself initializes unconditionally. It's a really basic fix.

**Comment by formatBCE** [comment]
> @formatBCE I'm not able to reproduce the crash on API 29 or 30, so not sure if this fix is correct or not. Are you?

Nope, i can't reproduce (but also I have no such old-ass devices, so it's emulator-only).

**Comment by seadowg** [comment]
> The registration sites (lines 508, 521) are already correctly guarded with SDK_INT < S checks. You just missed that the field itself initializes unconditionally. It's a really basic fix.

Agreed! The fix totally makes sense in theory, but I'm flummoxed by not being able to reproduce if it's correct. I can't recall ever having a problem with reproducing API level issues like this in the emulator. I also only have newer or too old devices.

**Comment by starbrightlab** [comment]
Meta portal dev has exploded in the last 2 weeks because they unlocked adb. I'm sat here looking at it working because I just built it from source with the fix. HA users are flooding to it.

**Comment by seadowg** [comment]
@starbrightlab OK cool! Let's run checks and then merge assuming everything is OK. I can continue to investigate why we're not able to reproduce without blocking.

File: composeApp/src/androidMain/kotlin/io/music_assistant/client/player/MediaPlayerController.android.kt [hunk 1]
```diff
@@ -83,7 +86,7 @@ actual class MediaPlayerController actual constructor(platformContext: PlatformC
         }
         // Resume is driven by AUDIOFOCUS_GAIN only — mode can return to NORMAL before
         // focus is restored, and we only auto-resume what we paused.
-    }
+    } }
```
seadowg: Formatting needs fixed here so it's:

```kotlin
private val modeChangedListener by lazy { 
    val inCall = mode == AudioManager.MODE_IN_CALL ||
            mode == AudioManager.MODE_IN_COMMUNICATION
    if (inCall && shouldPlayAudio) {
        logger.i { "Telephony mode=$mode — pausing server playback (focus backup)" }
        shouldPlayAudio = false
        pausedByFocusLoss = true
        audioTrack?.pause()
        onRemoteCommand?.invoke("pause")
    }
    // Resume is driven by AUDIOFOCUS_GAIN only — mode can return to NORMAL before
    // focus is restored, and we only auto-resume what we paused. 
}
```

That should fix lint.

seadowg (changes requested):

seadowg:
@starbrightlab @formatBCE got it! You need to use a release build type (so here either `release` or `selfSignedRelease`). It looks like the debug build helps us out with the class on older devices - that's still surprising to me, but makes some amount of sense. I was able to reproduce and validate that the fix works 👍.

seadowg (approved):

PR #657, opened 2026-06-19
Freeze playback position during seek/track-change until audio confirms

teancom:
This is the revamp of our progress bar and play/pause button for the local player that I mentioned in the Discord. I've been testing it for a couple of days and it seems to be *much* better. It has (virtually) eliminated the 'snap back' jumps when using the slider to jump around in a song, and the 'jittery' / bouncing when pausing or at the start of a track after track skip. This is *purely* for the local player, as I could use "do we actually have audio flowing to us right this second?" as a pretty good signal for whether we should start advancing the timer or not (for instance). And of course, for remote players we don't get that. That'll be a separate PR.

Commit message:

Local seeks and Next/Previous now freeze the progress bar at the target/boundary and resume only when the Sendspin stream actually syncs (confirmPlaying); server position updates are ignored while frozen, so the bar no longer ticks off the old stream or snaps on pre-audio queue updates. Also mask the play/pause button through the handoff (no pause->play->pause flicker), floor lock-screen seeks to whole seconds, and add a pendingPlay timeout backstop.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ iosApp/Frameworks/
 
 # Google Play developer-verification token (per-account, do not commit)
 androidApp/src/selfSignedRelease/assets/adi-registration.properties
+
+# Logs and other junk files
+tmp/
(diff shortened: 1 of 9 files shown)


formatBCE (commented):
Don't you feel like it's too complicated? I have a feeling that it could be simpler.

formatBCE (approved):

PR #658, opened 2026-06-19
Fixup some test names that do not compile on iOS

teancom:
Test names can't have commas in them, according to the iOS + kotlin build pipeline, so these break local tests.


formatBCE (approved):

formatBCE:
We should have Detekt rule for this

PR #659, opened 2026-06-20
Run iOS KMP tests in PR CI, fix tests

teancom:
We were only building in CI, tests were only being run during local development, for iOS specifically. Also, while I was in the neighborhood I fixed a long-standing bug that preventing running all tests at once, and added a check for test names that don't compile under Kotlin/Native.


teancom:
Oh, this was inspired by https://github.com/music-assistant/mobile-app/pull/658

formatBCE (approved):

Issue #589, opened 2026-06-10
Shortcuts


formatBCE:

seadowg:
Are you thinking about long press launcher shortcuts?

remon1496:
Also introduced in 2.9 :) Now allows for shortcuts in the MA frontend sidebar.

<img width="246" height="112" alt="Image" src="https://github.com/user-attachments/assets/b33a8d30-4657-43d0-905b-1ea9bbbd2b9c" />

formatBCE:
Yeah this. I think we can add it as a row to Home page. So user could have easy access to it should they want it.

seadowg:
Am I right in thinking these are purely a feature of the [frontend](https://github.com/music-assistant/frontend/)? It looks that way from skimming the code, and I can't see any API commands related to shortcuts.

formatBCE:
> Am I right in thinking these are purely a feature of the [frontend](https://github.com/music-assistant/frontend/)? It looks that way from skimming the code, and I can't see any API commands related to shortcuts.

Nope. It's using /me endpoint, sets shortcuts to user preference.

seadowg:
> It's using /me endpoint, sets shortcuts to user preference.

Ach! I'd misinterpreted the "preferences" code I was looking at. I see that now.

seadowg:
@formatBCE just reopening this: we still need to be able to edit shortcuts (add/remove them).

remon1496:
> We still need to be able to edit shortcuts (add/remove them).

How will adding/removing shortcuts work? Long-press menu? Handy to know for docs, prepping those.

seadowg:
@remon1496 yeah my thinking was an action in long press menu for grid items and top right options menu  for item details. I'm holding off for the moment while I discuss the API as the current direction we'd have to go with implementation (which I'm clarifying) feels like it'd be adding a lot of complexity to the mobile and already created web implementations we could probably offload on the server.

seadowg:
Adding the "Add to shortcut"/"Remove from shortcuts" button is currently blocked on future API changes that allow clients to receive an event when shortcuts are changed. It's been proposed that `MediaItem` will get a shortcut boolean property that will be updated as part of `media_item_updated`.

PR #661, opened 2026-06-20
Add shortcuts to Home screen

seadowg:
Work towards #589

Screenshot_20260620_141414

This adds shortcuts as a reorderable row (defaulting to the top) on the Home screen. Adding/removing shortcuts can be a follow-up - this lays enough foundation for a review and is useful on its own.

TODO:
- [x] Tidy up shortcut `ItemCategory` hardcoded values
- [x] Work out if we can decode `sidebar.shortcuts` key with `@Serializable` classes (if not, we can feedback about this to server team)
- [x] Work out if there's a better way to grab multiple items via URIs than with multiple `item_by_uri` commands

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/feature/HomeTest.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/feature/HomeTest.kt
@@ -27,7 +27,7 @@ class HomeTest {
     val serviceClient: FakeServiceClient by inject(ServiceClient::class.java)
 
     @Test
-    fun `can refresh home`() {
+    fun `can refresh home recommendations`() {
         val album1 = ServerMediaItemFixtures.album()
         serviceClient.addToLibrary(album1)
 
@@ -39,5 +39,18 @@ class HomeTest {
 
         homePage.refresh()
             .assertMediaDisplayed(album2.name)
+    }
+
+    @Test
+    fun `can refresh home shortcuts`() {
+        val album = ServerMediaItemFixtures.album()
+        serviceClient.addToLibrary(album)
+
+        val homePage = launchLoggedInApp(composeTestRule, serviceClient)
+
+        serviceClient.addShortcut(album)
+
+        homePage.refresh()
+            .assertShortcutDisplayed(album)
     }
 }
--- a/androidApp/src/test/kotlin/io/music_assistant/client/feature/HomeTest.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/feature/HomeTest.kt
@@ -8,6 +8,7 @@ import io.music_assistant.client.support.Qualifiers
 import io.music_assistant.client.support.ServerMediaItemFixtures
 import io.music_assistant.client.support.launchLoggedInApp
 import io.music_assistant.client.support.pages.assertMediaDisplayed
+import io.music_assistant.client.support.pages.assertMediaNotDisplayed
 import io.music_assistant.client.support.rules.createTestRuleChain
 import org.junit.Rule
 import org.junit.Test
@@ -52,5 +53,21 @@ class HomeTest {
 
         homePage.refresh()
             .assertShortcutDisplayed(album)
+    }
+
+    @Test
+    fun `shows error if data can't be loaded and can recover with refresh`() {
+        val album = ServerMediaItemFixtures.album()
+        serviceClient.addToLibrary(album)
+        val homePage = launchLoggedInApp(composeTestRule, serviceClient)
+
+        serviceClient.setRequestErrors(true)
+        homePage.refresh()
+            .assertErrorLoadingData()
+            .assertMediaNotDisplayed(album.name)
+
+        serviceClient.setRequestErrors(false)
+        homePage.refresh()
+            .assertMediaDisplayed(album.name)
     }
 }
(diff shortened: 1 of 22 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/GridButton.kt
```diff
@@ -0,0 +1,54 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun GridButton(
```
seadowg, line 22: This ended up being unrelated, but still a nice change to have I reckon?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -177,19 +184,54 @@ class HomeScreenViewModel(
         }
     }
 
-    fun loadRecommendations(): Job = viewModelScope.launch {
-        _recommendationsState.update { it.copy(recommendations = DataState.Loading()) }
-        repeat(MAX_RECOMMENDATION_ATTEMPTS) { attempt ->
```
seadowg, line 182: I don't **think** we need this any longer now that we have manual refreshes - open to hearing about that not being correct though! I've added an explicit error state as well so that the user isn't stuck with an infinite loading screen.

androidApp/src/test/kotlin/io/music_assistant/client/feature/HomeTest.kt
```diff
@@ -40,4 +41,33 @@ class HomeTest {
...
+        val homePage = launchLoggedInApp(composeTestRule, serviceClient)
+
+        serviceClient.addShortcut(album)
+
+        homePage.refresh()
+            .assertShortcutDisplayed(album)
+    }
+
+    @Test
+    fun `shows error if data can't be loaded and can recover with refresh`() {
```
seadowg, line 59: I might have a look at backfilling more error state tests after this as it'd be good to harden our resilience around regressions for these.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DisplayString.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import androidx.compose.runtime.Composable
+import org.jetbrains.compose.resources.StringResource
+import org.jetbrains.compose.resources.stringResource
+
+/**
+ * Abstraction that allows params/properties to use [StringResource] or [String] without the actual
+ * underlying type needing to be known.
+ */
+sealed class DisplayString {
```
seadowg, line 11: This is a utility I end up creating in every Android project. If there's something existing that could be used instead, please shout at me!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreen.kt
```diff
@@ -279,6 +247,60 @@ fun HomeScreen(
...
+            recommendations
+        }
+    } else {
+        emptyList()
+    }
+
+    return reconcileHomeRows(baseList, homeRowsConfig, onTop = SHORTCUTS_CATEGORY_ID)
+}
+
+private const val SHORTCUTS_CATEGORY_ID = "shortcuts"
```
seadowg, line 302: @formatBCE I can't find any docs on reserved or unreserved forms of `itemId`, so there's definitely a danger this could clash with a server provided `shortcuts` `RecommendationFolder` in the future. Do you think we should attempt to avoid that with a prefix or do you have a better idea of a form we could use that would be safer?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -177,19 +183,50 @@ class HomeScreenViewModel(
...
+            it.copy(
+                recommendations = DataState.Loading(),
+                shortcuts = DataState.Loading(),
             )
-                ?.let { items ->
-                    _recommendationsState.update { it.copy(recommendations = DataState.Data(items)) }
-                    return@launch
+        }
+
+        loadDataJob = viewModelScope.launch {
```
seadowg, line 196: @formatBCE I've noticed that all the `sendRequest` calls happen in `launch` blocks with the default `Dispatcher`. I'm assuming the websockets/webRTC implementations then execute the actual network IO off of the UI thread and that's just abstracted away at this level?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DisplayString.kt
formatBCE: Yup, same :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
formatBCE: Need to check that (many refactorings came and go after initial implementation), but whole network stack was using IO.

formatBCE (approved):

PR #667, opened 2026-06-22
Add shadow under `FloatingBar`

seadowg:
This was an idea I never got to when we first introduced the `FloatingBar`. Looking at @formatBCE's status bar shadow reminded me about it (and gave me homework to copy).

The effect is very subtle, but I think it's a nice touch!


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/FloatingBar.kt
```diff
@@ -104,6 +117,22 @@ fun FloatingBarLayout(
...
+@Composable
+private fun FloatingBarShadow(modifier: Modifier) {
+    Box(
+        modifier = modifier
+            .fillMaxWidth()
+            .background(
+                Brush.verticalGradient(
+                    listOf(
+                        Color.Transparent,
+                        Color.Black.copy(alpha = 0.6f),
```
seadowg, line 129: I experimented with using theme surface colors, but found just using black with this alpha worked well in both the dark and light themes.

formatBCE (approved):

PR #685, opened 2026-06-25
feat: iOS lock-screen and media widget skip back/forward for long-form content

teancom:
This updates the iOS lock screen and media widget to use the 10 second/30 second buttons when we're playing back audiobooks or podcasts, that we recently added in-app. I centralized the values to support (in the future) letting users override the values if they want to, but didn't want to bundle UI / new configs into the same PR. I also didn't touch the Android side any more than necessary to keep it compiling, but it should be easy enough to do the same there.

Commit message:

Adds audiobook/podcast-aware skip buttons to the iOS lock screen:
- LongFormSeekDefaults centralizes 10s back / 30s forward intervals
- seek_by: protocol added to LocalPlayerController for relative seeks
- iOS NowPlayingManager resolves skip_back/skip_forward to seek_by:±N and toggles prev/next vs skip back/forward based on content type
- Seek intervals flow through a dedicated setLongFormSeekIntervals channel rather than riding every updateNowPlaying call, so the per-second now-playing path no longer touches remote-command state
- Android and in-app controls use the same shared constants


formatBCE (approved):
We will have to revise splitting logic somehow. It becomes pretty inconvenient to read.

PR #688, opened 2026-06-26
Fix LazyGrid/List duplicate-key crashes as a class

teancom:
I was going through the Test Flight submitted crash reports and found that there were still some coming in after our last round of LazyList duplicate-key crash fixes in c9447a7. Apparently having duplicate items served to us from the server is more common than I thought, in more areas. This is an attempt to prevent those crashes everywhere we use `Lazy*` layouts without causing a bunch of re-ordering or performance issues. 🤞

Commit message:

Compose aborts (requirePrecondition / "Key was already used" in SubcomposeLayout) when a lazy layout is handed duplicate keys, which is the dominant TestFlight crash signature on fling/prefetch. Keying media lists by canonical identity alone collides whenever a collection legitimately repeats an item (recommendations, search results, playlists, folders, queues).

- Add lazyListOccurrenceKeys(): pairs each canonical key with a per-key occurrence ordinal so duplicates stay unique without churning unrelated items' keys on insert/remove.
- Convert the at-risk lazy layouts to itemsIndexed + occurrence keys: LibraryScreen, AdaptiveMediaGrid, ItemDetailsScreen, CategoryRow (recommendations), and SearchScreen.
- Give ItemDetailsScreen's static grid sections explicit stable keys.
- Pin the key contract with LazyListKeysTest.

Verified: testAndroidHostTest, iosSimulatorArm64Test, and compileTestKotlinIosArm64 all pass.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
@@ -1,57 +1,47 @@
 package io.music_assistant.client.ui.compose.common.items
 
-import io.music_assistant.client.data.model.client.MediaType
 import io.music_assistant.client.data.model.client.items.AppMediaItem
 import io.music_assistant.client.data.model.client.items.PlayableItem
-import kotlin.jvm.JvmName
 
 /**
- * Stable canonical identity for an [AppMediaItem] inside a Compose `LazyColumn`,
- * `LazyRow`, or `LazyVerticalGrid`'s `key = { ... }` lambda when the rendered
- * collection cannot contain the same canonical item twice.
+ * Stable canonical identity for an [AppMediaItem] as a single collision-safe
+ * [String], combining `mediaType`, `provider`, and `itemId`.
  *
- * Combines `mediaType`, `provider`, and `itemId` so the same canonical
- * `itemId` carried by two different providers (e.g. a track with the same
- * server-side `itemId` from Spotify and Apple Music) produces distinct
- * keys. Without this, Compose's `SubcomposeLayout.subcompose` precondition
- * (`SubcomposeLayout.kt:614` — "Key X was already used") fires during
- * fling/prefetch on iOS and the app crashes.
+ * Use this as a Compose lazy-layout `key` ONLY when the rendered collection
+ * cannot contain the same canonical item twice. For collections that may
+ * legitimately repeat an item (folders, playlists, queues, search results),
+ * use [lazyListOccurrenceKeys] instead.
  *
- * Returns a structured [Triple] rather than a delimited string so the
- * encoding cannot accidentally collide on field contents — e.g. without
- * this, `(provider="apple_music", itemId="xyz")` would generate the same
- * underscore-joined string as `(provider="apple", itemId="music_xyz")`.
- * Compose's slot identity uses `equals`/`hashCode`, both of which Triple
- * implements structurally.
- *
- * Contract pinned by `LazyListKeysTest`.
+ * The encoding is length-prefixed so field contents can never collide across
+ * field boundaries — e.g. (provider="apple_music", itemId="xyz") is distinct
+ * from (provider="apple", itemId="music_xyz").
  */
-fun AppMediaItem.lazyListKey(): Triple<MediaType, String, String> =
-    Triple(mediaType, provider, itemId)
+fun AppMediaItem.lazyListKey(): String =
+    mediaItemKey(mediaType.name, provider, itemId)
 
 /**
  * Keys a lazy-layout list of media items that may legitimately contain the same
- * canonical [lazyListKey] more than once, such as folders, playlists, queues, or
+ * canonical media item more than once, such as folders, playlists, queues, or
  * other occurrence-based server data.
  */
-fun List<AppMediaItem>.lazyListOccurrenceKeys(): List<Pair<Triple<MediaType, String, String>, Int>> =
+fun List<AppMediaItem>.lazyListOccurrenceKeys(): List<String> =
     occurrenceKeys { it.lazyListKey() }
 
 /**
- * [PlayableItem] does not expose [AppMediaItem.mediaType], but every concrete
- * [PlayableItem] in this codebase is also an [AppMediaItem], so each item is
- * narrowed here to reuse the canonical [lazyListKey].
+ * Keys playable lists with the same canonical identity as [AppMediaItem].
  */
-@JvmName("lazyListOccurrenceKeysPlayable")
-fun List<PlayableItem>.lazyListOccurrenceKeys(): List<Pair<Triple<MediaType, String, String>, Int>> =
-    occurrenceKeys { (it as AppMediaItem).lazyListKey() }
+fun List<PlayableItem>.playableLazyListOccurrenceKeys(): List<String> =
+    occurrenceKeys { item -> mediaItemKey(item.mediaType.name, item.provider, item.itemId) }
 
-private fun <T, K> List<T>.occurrenceKeys(keyOf: (T) -> K): List<Pair<K, Int>> {
-    val occurrences = mutableMapOf<K, Int>()
+private fun <T> List<T>.occurrenceKeys(keyOf: (T) -> String): List<String> {
+    val occurrences = mutableMapOf<String, Int>()
     return map { item ->
         val key = keyOf(item)
         val occurrence = occurrences.getOrElse(key) { 0 }
         occurrences[key] = occurrence + 1
-        key to occurrence
+        if (occurrence == 0) key else mediaItemKey("occurrence", key, occurrence.toString())
     }
 }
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
```diff
@@ -25,3 +28,30 @@ import io.music_assistant.client.data.model.client.items.AppMediaItem
...
+ */
+fun List<AppMediaItem>.lazyListOccurrenceKeys(): List<Pair<Triple<MediaType, String, String>, Int>> =
+    occurrenceKeys { it.lazyListKey() }
+
+/**
+ * [PlayableItem] does not expose [AppMediaItem.mediaType], but every concrete
+ * [PlayableItem] in this codebase is also an [AppMediaItem], so each item is
+ * narrowed here to reuse the canonical [lazyListKey].
+ */
+@JvmName("lazyListOccurrenceKeysPlayable")
```
formatBCE: Why this needed?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
```diff
@@ -25,3 +28,30 @@ import io.music_assistant.client.data.model.client.items.AppMediaItem
...
+fun List<PlayableItem>.lazyListOccurrenceKeys(): List<Pair<Triple<MediaType, String, String>, Int>> =
+    occurrenceKeys { (it as AppMediaItem).lazyListKey() }
```
formatBCE: This is asking for troubles. There's no shielding against someone implementing some PlayableItem in class that doesn't extend AppMediaItem. Use `as?` and guard against those.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
```diff
@@ -25,3 +28,30 @@ import io.music_assistant.client.data.model.client.items.AppMediaItem
  */
 fun AppMediaItem.lazyListKey(): Triple<MediaType, String, String> =
     Triple(mediaType, provider, itemId)
+
+/**
+ * Keys a lazy-layout list of media items that may legitimately contain the same
+ * canonical [lazyListKey] more than once, such as folders, playlists, queues, or
+ * other occurrence-based server data.
+ */
+fun List<AppMediaItem>.lazyListOccurrenceKeys(): List<Pair<Triple<MediaType, String, String>, Int>> =
```
formatBCE: this `List<Pair<Triple<MediaType, String, String>, Int>>` looks not good. :) 
I propose at least replacing it with some data class with 4 fields. The best way would be to use compound string as a key, as strings are final and have good heap caching, so don't use memory every time they're created. Same actually would stands for that triples we have in class. It's constant objects re-creation on every list refresh, which leads to garbage collections and stuttering - not to mention semantic complexity.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -689,8 +696,20 @@ private fun DetailGrid(
     }
 }
 
-private fun LazyGridScope.fullSpanItem(content: @Composable () -> Unit) =
-    item(span = { GridItemSpan(maxLineSpan) }) { content() }
+private fun LazyGridScope.fullSpanItem(
```
formatBCE, line 694: Are there occurrences of getting two identical rows here?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -650,11 +651,17 @@ private fun LazyGridScope.detailHeaderItems(
     heroSlot: @Composable () -> Unit,
     tabsSlot: (@Composable () -> Unit)?,
 ) {
-    item(span = { GridItemSpan(maxLineSpan) }) {
+    item(
```
formatBCE: Use `fullSpanItem` everywhere?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -650,11 +651,17 @@ private fun LazyGridScope.detailHeaderItems(
...
+        key = DETAIL_HERO_KEY,
+        span = { GridItemSpan(maxLineSpan) },
+    ) {
         Box(modifier = Modifier.fullBleed(gridPadding)) { heroSlot() }
     }
     tabsSlot?.let { slot ->
-        item(span = { GridItemSpan(maxLineSpan) }) { slot() }
+        item(
```
formatBCE: Same.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -689,8 +696,20 @@ private fun DetailGrid(
...
-    item(span = { GridItemSpan(maxLineSpan) }) { content() }
+private fun LazyGridScope.fullSpanItem(
+    key: String,
+    content: @Composable () -> Unit,
+) = item(
+    key = key,
+    span = { GridItemSpan(maxLineSpan) },
+) { content() }
+
+private const val DETAIL_HERO_KEY = "detail:hero"
```
formatBCE: These constants in the middle of code - can we move them to the bottom?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
teancom: Yeah, this was the site of the actual crashes that inspired this branch. The rest was me trying to fix these "for good" vs just the one place that we were dying (given this was the second time I was investigating Lazy* crashes, I didn't want to do another partial fix).

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -780,16 +802,18 @@ private fun ArtistsTabContent(
...
         tabListBody(artistsState) { artists ->
-            items(
+            val artistKeys = artists.lazyListOccurrenceKeys()
+            itemsIndexed(
                 items = artists,
+                key = { index, _ -> artistKeys[index] },
                 span = when (viewMode) {
                     ViewMode.LIST -> {
-                        { GridItemSpan(maxLineSpan) }
+                        { _, _ -> GridItemSpan(maxLineSpan) }
```
formatBCE: I believe you can omit these `_, _ ->`

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
```diff
@@ -25,3 +28,30 @@ import io.music_assistant.client.data.model.client.items.AppMediaItem
  */
 fun AppMediaItem.lazyListKey(): Triple<MediaType, String, String> =
```
formatBCE: I believe we could replace every occurrence of this one with new method, and make it private of inline it?

formatBCE (changes requested):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
teancom: Not within a when/if branch, it doesn't compile because the lambda literal doesn't have an expected type.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/LazyListKeys.kt
teancom: I'm overloading lazyListOccurrenceKeys for both AppMediaItem and PlayableItem. They're sibling types so a single overload can't serve both. But then, the JVM/Android compiler erases generics and the function signature clashes (kotlin/native doesn't do things the same way so it works without the annotation there).

formatBCE: If this happens, I think we need to do it other way. Like, for example, exposing media type from playable item.

teancom:
@formatBCE Ready for a re-review. Thank you!

formatBCE (approved):
Thank you for changes! Looks good!

Issue #412, opened 2026-05-18
Large screen landscape mode left menu bar joins late


remon1496:
## Context
 - Operating system and version: iPadOS 26.5
 - App version: TestFlight 1.0 (27)
 - Music Assistant server version: 2.8.7 Stable

### Problem description

When in landscape mode switching screens, there is a slide animation to the next or previous screen. The sidebar/navigation bar joins after the animation in a next step on the navigational stack. Precious page animation feels good because it naturally flows in from the left side.

### Steps to reproduce

1. Have iPad in landscape mode
2. Navigate over different pages

### Expected behavior

I think it would feel most natural if the sidebar/menubar slides connected to the screen its in, instead of popping in from the left after the transition of the page is done.

https://github.com/user-attachments/assets/ecd72a90-5cb1-4d4e-9e01-67255c7d2f99

remon1496:
@formatBCE I feel like this is resolved (at least on android) in the latest build. Did we move from sliding animations to fading animations? Can you confirm?

formatBCE:
There were no explicit animations at all. :) Not sure what has changed.

remon1496:
> There were no explicit animations at all. :) Not sure what has changed.

Well, we'll see on next iOS build :) I'll close this one when confirmed.

Robot tells me this:

> Yes, out-of-the-box UI animations in Kotlin Multiplatform (KMP/Compose) are different from Android. When you use shared Compose UI, the framework intentionally maps animations to feel native to the device they are running on.On iOS: Screens slide in from the right when pushed and to the left when popped (mirroring SwiftUI), and iOS back edge swipe gestures are supported by default.On Android: Navigation default animations typically use fading or simpler transitions, respecting native Android Predictive Back gestures.

remon1496:
Resolved in TestFlight 1.0 (29). Likely fixed by https://github.com/music-assistant/mobile-app/pull/433.

seadowg:
Reopening this as we've re-enabled transition animations with #691. This will need a proper fix that allows the default iOS transition animations to work with the side nav bar.

PR #691, opened 2026-06-27
Restore main nav gestures

### LINKED ISSUE #284: Bug: App crashes on several interaction when using 'Reduce Motion'
I have my phone to use reduced motion, this causes some interactions to crash the app.
The interactions I've seen so far that cause it to crash (haven't gone through all possible actions):
- Select any player from the player list
- Clicking the Done button on the Group settings
- Selecting a previously saved server to connect to

Steps to reproduce:
- Enable 'Reduced motion' (in iOS: Settings > Accessibility > Motion > Reduce Motion)
- Open the app
- Perform any of the actions described above

I assume Android also has options for reduced motion, I don't know if Android also crashes with it enabled, I don't have an android device available to test.

### LINKED ISSUE #675: Predictive Back Gesture is missing on iOS
Hi,

As requested here is the report of the missing back gesture on iOS.

Compare the back gesture with other big apps like WhatsApp how it should look. 

Best regards

seadowg:
Closes https://github.com/music-assistant/mobile-app/issues/675

It looks like we don't need the workaround that caused this issue any longer as we've upgraded to a new version of KMP that appears to fix https://github.com/music-assistant/mobile-app/issues/284 (I couldn't reproduce in the simulator with these changes).

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
@@ -2,9 +2,6 @@
 
 package io.music_assistant.client.ui.compose.home
 
-import androidx.compose.animation.EnterTransition
-import androidx.compose.animation.ExitTransition
-import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
@@ -295,13 +292,6 @@ fun MainNavigationRoot(
                             multiBackStack.removeLastOrNull()
                         },
                         backEnabled = !playerExpanded,
-                        // Workaround for CMP 1.10.3 iOS crash: LazyLayout measured inside
-                        // AnimatedContent + CupertinoOverscroll trips a SubcomposeLayout
-                        // precondition on first frame. Disabling transitions removes the
-                        // animating measure path.
-                        transitionSpec = { EnterTransition.None togetherWith ExitTransition.None },
-                        popTransitionSpec = { EnterTransition.None togetherWith ExitTransition.None },
-                        predictivePopTransitionSpec = { EnterTransition.None togetherWith ExitTransition.None },
                     )
                 }
             }


formatBCE:
But it was crash fix.

Any replacement? :)

P.S. sorry, description didn't load at first...

Well, let's try. I'm pessimistic.

formatBCE (approved):

seadowg:
> Well, let's try. I'm pessimistic.

@formatBCE we can revert the PR directly if it causes a regression. @remon1496 can you make sure to include a note in the the next TestFlight notes about this? Let me know if you need more details.

remon1496:
> make sure to include a note in the the next TestFlight notes about this?

Will do!

remon1496:
Did you also test on an virtual iPad @seadowg? Because of this thingy right here: https://github.com/music-assistant/mobile-app/issues/412

seadowg:
@remon1496 no I didn't! Thanks for prompting on that. It does look like these changes do indeed reintroduce that issue. We can revert this PR, or reopen https://github.com/music-assistant/mobile-app/issues/412 and find a different fix for it. Which would you prefer @remon1496? 

Personally, I'd opt to revert this - preditive back is still pretty rare on Android and as much as it's nice, it's not something that I've gotten used to. I don't really have a feel for how ingrained it's become on iOS.

remon1496:
@seadowg Well, on iOS it's the default (all system default apps offer this), and a lot of third-party apps also have it. So for the sake of the platform, I would prefer to have it. Would it bother me to not have it? Not really tbh, but I could imagine other users would.

Possible way to solve #412 is maybe to 'stick' the left menu/sidebar to the rest of the view so they animate alongside each other?

seadowg:
> Well, on iOS it's the default (all system default apps offer this), and a lot of third-party apps also have it. So for the sake of the platform, I would prefer to have it.

Ok let's reopen #412 for now then!

> Possible way to solve https://github.com/music-assistant/mobile-app/issues/412 is maybe to 'stick' the left menu/sidebar to the rest of the view so they animate alongside each other?

That's kind of how it works right now - the animation or the layout order just needs to be tweaked so that it works with the iOS "slide in" transition animation. KMP gives us different default animations for Android/iOS. That's really nice, but just means that we don't see this problem on Android tablets as they use a different animation which doesn't mess with the side nav bar.

PR #696, opened 2026-06-28
Fix iOS transitions with side bar

### LINKED ISSUE #412: Large screen landscape mode left menu bar joins late

seadowg:
Closes https://github.com/music-assistant/mobile-app/issues/412

This simplifies our adaptive navigation layout so that we take advantage of the fixed height of both the rail and bar and show content underneath them (with content padding passed to it). This means that transitions show underneath these bars which allowing for things like slide in/out to look more natural and not mess with nav controls.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationBarLayout.kt
```diff
@@ -30,53 +33,59 @@ import io.music_assistant.client.utils.WindowClass
...
     navigationItems: List<NavigationItem>,
-    showNavBar: Boolean = true,
+    showNavigation: Boolean = true,
+    navigationBarHeight: Dp = 88.dp,
+    navigationRailWidth: Dp = 80.dp,
     content: @Composable BoxScope.(contentPadding: PaddingValues) -> Unit,
 ) {
     val isExpandedScreen = WindowClass.isAtLeastExpanded()
 
-    Scaffold(
```
seadowg, line 40: We don't need a `Scaffold` here any longer: the nav controls are both fixed size, so there's no need for measuring passes to get content padding.

seadowg:
@teancom I've tested this on an iPhone and iPad simulator, but would appreciate a double check that everything looks correct here!

teancom:
> @teancom I've tested this on an iPhone and iPad simulator, but would appreciate a double check that everything looks correct here!

I'll trade you - I'll look at this if you take a look at https://github.com/music-assistant/mobile-app/pull/697 😄

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationBarLayout.kt
```diff
@@ -135,5 +147,5 @@ fun PreviewAdaptiveNavigationScaffold() {
 )
 @Composable
 fun PreviewAdaptiveNavigationScaffoldExpanded() {
```
teancom: Should probably change 'Scaffold' in this function name as well?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/AdaptiveNavigationBarLayout.kt
```diff
@@ -30,53 +33,59 @@ import io.music_assistant.client.utils.WindowClass
...
-                            },
-                        )
-                    }
+        } else if (showBar) {
+            NavigationBar(
+                modifier = Modifier.align(Alignment.BottomCenter).height(navigationBarHeight),
+                containerColor = MaterialTheme.colorScheme.surfaceContainerHigh,
+            ) {
+                navigationItems.forEach {
+                    NavigationBarItem(
```
teancom, line 80: We're using named arguments here and positional arguments for NavigationRailItem just above. This was already true, but we could align them while we're here, if you want (not blocking on this).

teancom (approved):
Coupla tiny nits but it works great on real hardware, iPhone and iPad Pro. Thank you!

PR #697, opened 2026-06-28
Warn when hardware volume buttons are used on a remote player

### LINKED ISSUE #621: Hint when trying to use volume controls for players
## Problem

When viewing the expanded player, we've heard that some users expect the volume controls on their device to control the volume of the player they're currently looking at. This does "work" for the local player as the changing device volume will change the volume of playback.

## Solution

Instead of taking control of the device volume controls (which can be very confusing) we could show a hint to the user if they try and use the volume controls while the expanded player is visible and non-local player is selected. The simplest solution is probably just to show a Toast (via `ToastHost`) with a short sentence prompting the user to use the on-screen volume control for the player.

teancom:
Show a toast when the user presses the hardware volume buttons while viewing/controlling a non-local player, since those buttons only affect the local (on-device) player. Local-player volume continues to work silently with no toast.

- Android: MainActivity.dispatchKeyEvent emits a RemoteVolumeButtonEvent on volume up/down (still passing the event through to the system).
- iOS: SystemVolumeButtonObserver observes AVAudioSession.outputVolume via KVO. It only activates the shared session when the local engine is not rendering, and does so with .mixWithOthers so observing for a remote player no longer interrupts other audio apps (e.g. Qobuz). NowPlayingManager.activatePlayback() re-asserts exclusive non-mixing .playback so local playback always reclaims focus.
- Common: PlatformVolumeButtonObserver bridge + RemoteVolumeButtonEvents flow; MainNavigationRoot starts/stops the observer and shows the toast based on the viewed player's isLocal.
- ToastHost now respects the bottom safe-area inset.

Closes https://github.com/music-assistant/mobile-app/issues/621


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -197,6 +222,32 @@ fun MainNavigationRoot(
     val libraryScreenState = LibraryScreenState.create()
     val searchScreenState = SearchScreenState.create()
 
+    val remoteVolumeHint = stringResource(Res.string.players_remote_volume_hint)
+    val viewedPlayer = data?.playerData?.getOrNull(playerPagerState.settledPage)
+        ?: data?.selectedPlayerIndex?.let { data.playerData.getOrNull(it) }
+    val shouldObservePlatformVolumeButtons = viewedPlayer?.isLocal == false
```
formatBCE: `viewedPlayer?.isLocal != true`?

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -83,13 +89,32 @@ import musicassistantclient.composeapp.generated.resources.nav_home
...
 import org.koin.core.parameter.parametersOf
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
+interface PlatformVolumeButtonObserver {
+    fun start()
+    fun stop()
+}
+
+object PlatformVolumeButtonObserverProvider {
```
formatBCE: These objects are basically singletones, acting like some kind of event bus - but it's anti-pattern. We want to use proper dependency injection in places where it's needed - especially dangerous is this one, with public var, but `RemoteVolumeButtonEvents` is also not good.

androidApp/src/main/kotlin/io/music_assistant/client/MainActivity.kt
```diff
@@ -73,6 +76,20 @@ class MainActivity : ComponentActivity() {
         handleIncomingUri(intent)
     }
 
+    // ComponentActivity.dispatchKeyEvent carries a @RestrictTo(LIBRARY_GROUP_PREFIX)
+    // annotation, so calling super trips the RestrictedApi lint. Overriding it to
+    // observe hardware volume keys is a legitimate, supported use; we still forward
+    // the event to super so the system handles volume normally.
+    @SuppressLint("RestrictedApi")
+    override fun dispatchKeyEvent(event: KeyEvent): Boolean {
```
formatBCE: If you use `onKeyDown`, you won't need restricted API suppression:
```
override fun onKeyDown(keyCode: Int, event: KeyEvent?): Boolean {
        if (keyCode == KeyEvent.KEYCODE_VOLUME_UP || keyCode == KeyEvent.KEYCODE_VOLUME_DOWN) {
            RemoteVolumeButtonEvents.emit()
        }
        return super.onKeyDown(keyCode, event)
    }
```

formatBCE:
Overall, I would like to keep unharnessed agentic code out from core. It is helpful for sure - but it often introduces sub-optimal decisions, that bring uncertainty and tech debt. :)
Not to diminish the work you've done @teancom ! It's just there are me and @seadowg that can hold agent's horses a bit and don't let it go nuts. :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
teancom: I had it this way to default to 'do nothing' if we weren't sure (still loading, whatever) if the observed player was local or remote. I *can* reverse it but that defaults to "show the toast if we aren't sure", and I figured the safer default would be "don't show".

formatBCE: Well, it's informational toast, so whatever. Keep it :)

formatBCE (approved):
Great! Thanks for refactoring!

teancom:
@formatBCE I actually found a bug (or at least an annoying issue) where volume changes while the app was in the background would get 'marked' and then the toast would pop up the next time the app was foregrounded if you were on a remote player, even if it was much later. I'm fixing that and cleaning it up some more, and then will test some more. I should have marked this as draft as soon as I found the bug a few hours ago, sorry.

seadowg (changes requested):
Just marking request changes while it's a draft and you can re-request when it's ready to go!

formatBCE:
I don't really like those booleans that agents put everywhere...
Let me please do one take on it @teancom - I will ask you to check your background bug after my commit.

teancom:
@formatBCE Hell yeah. If you want to use any of the code, feel free to push to this branch. Or open a new one. Either way, I'd love to learn.

formatBCE:
> @formatBCE Hell yeah. If you want to use any of the code, feel free to push to this branch. Or open a new one. Either way, I'd love to learn.

Check it out. I didn't delete comments there so it's more obvious what and why.
I know now that iOS is pretty bad at this, but it should be harnessed still. Please verify on live device.

teancom:
Works great, doesn't have either of the bugs that I ran into before.

formatBCE (approved):

teancom:
@seadowg This is ready for re-review whenever you have a chance. Also, do I need to change the title? Do we want/care about `feat:` or `fix:` or whatever at the start, when it comes to the changelog?

seadowg:
> Also, do I need to change the title? Do we want/care about feat: or fix: or whatever at the start, when it comes to the changelog?

Honestly I'd prefer to lose it. I can update it.

composeApp/src/commonMain/kotlin/io/music_assistant/client/input/RemoteVolumeButtonController.kt
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Platform hook for volume buttons that need active observation (iOS KVO).
+ * [start]/[stop] must be idempotent — repeated calls in the same state are no-ops.
+ */
+interface PlatformVolumeButtonObserver {
+    fun start()
+    fun stop()
+}
+
+class RemoteVolumeButtonController {
```
seadowg: I would just call this `VolumeButtonService`. It doesn't really have any features that would make it a "controller" in any traditional sense - especially since we're interacting with it from iOS code where there's a very specific application of the pattern.

composeApp/src/commonMain/kotlin/io/music_assistant/client/input/RemoteVolumeButtonController.kt
```diff
@@ -0,0 +1,42 @@
...
+     * and the (expensive) iOS observer, which flips the shared audio session to
+     * `.mixWithOthers` and so must only run while a remote player is on screen.
+     */
+    var observingRemote: Boolean = false
+        set(value) {
+            field = value
+            if (value) platformObserver?.start() else platformObserver?.stop()
+        }
+
+    fun setPlatformObserver(observer: PlatformVolumeButtonObserver?) {
```
seadowg: I like the general approach here, but I'm not loving that we have this shared component for abstracting volume button interactions, but it ends up with a platform specific API (Android doesn't use the `PlatformObserver`). It also feels like there's a lot of lifecycle to manage here. A couple of thoughts on this:

- Could the iOS code just manage the `SystemVolumeButtonObserver` itself and activate/deactivate the observation based on if the app is on screen or not? I believe that in the Android case `onKeyEvent` isn't called if the Activity is in the background, so even the lifecycle handling in `MainNavRoot` is platform specific here.
- If we can just have the `SystemVolumeButtonObserver` lifecycle manged in iOS code, I think the UI code in `MainNavRoot` could be simpler.
- This interaction be tested at a UI level with a faked `RemoteVolumeButtonController` that can be triggered from the tests.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -198,6 +206,29 @@ fun MainNavigationRoot(
     val libraryScreenState = LibraryScreenState.create()
     val searchScreenState = SearchScreenState.create()
 
+    val remoteVolumeHint = stringResource(Res.string.players_remote_volume_hint)
+    val viewedPlayer = data?.playerData?.getOrNull(playerPagerState.settledPage)
```
seadowg: You can grab the current player from the `PlayerState.Data` without needing to care about the pager state with something like this:

```
val viewedPlayer = if (data != null && data.selectedPlayerIndex != null) {
    data.playerData[data.selectedPlayerIndex]
} else {
    null
}
```

Maybe we could add that as an extension function?

seadowg (changes requested):

teancom:
@seadowg Ready for a re-review whenever you have a chance. 🙇

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -207,24 +206,15 @@ fun MainNavigationRoot(
...
-    }
-
-    // Collect only while the UI is on screen; off-screen presses land on no
-    // subscriber and are dropped (no replay on return). The hint is read through
-    // rememberUpdatedState so a locale change is reflected without restarting.
+    val viewingRemote = data?.selectedPlayer?.isLocal == false
     val currentHint by rememberUpdatedState(remoteVolumeHint)
     val lifecycleOwner = LocalLifecycleOwner.current
-    LaunchedEffect(lifecycleOwner, remoteVolumeButtonController) {
+    LaunchedEffect(lifecycleOwner, volumeButtonService, viewingRemote) {
```
formatBCE: Why "viewingRemote in parameter? This will cancel and re-subscribe every time that value is changed. But we don't really need to re-establish collection...

PR #719, opened 2026-07-01
sidebar.shortcuts are not guaranteed to exist

teancom:

Not really, just noticed that we were expecting them to always exist and I don't have them, so it logged an ugly RPC decoding failure.

```
kotlinx.serialization.MissingFieldException: Field 'sidebar.shortcuts' is required for type with serial name 'io.music_assistant.client.data.model.server.ServerUserPreferences', but it was missing
```
(and 52 other lines).

Not an 'additional change' per se, but I'm not sure this sort of commit needs to go into the change-log?

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerUser.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerUser.kt
@@ -10,5 +10,5 @@ data class ServerUser(
 
 @Serializable
 data class ServerUserPreferences(
-    @SerialName("sidebar.shortcuts") val shortcuts: List<String>,
+    @SerialName("sidebar.shortcuts") val shortcuts: List<String> = emptyList(),
 )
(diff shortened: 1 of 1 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerUser.kt
```diff
@@ -10,5 +10,5 @@ data class ServerUser(
 
 @Serializable
 data class ServerUserPreferences(
-    @SerialName("sidebar.shortcuts") val shortcuts: List<String>,
+    @SerialName("sidebar.shortcuts") val shortcuts: List<String> = emptyList(),
```
formatBCE: I like `List<String>? = null`, and then deal with nullability on site. Gives you knowledge that there was no field versus having empty one (and saves tiny bit of heap RAM). :)

formatBCE (approved):

PR #723, opened 2026-07-01
Clean up detekt host test findings

teancom:

I ran `./gradlew :composeApp:detektAndroidHostTest` and discovered that there were a *ton* of findings. This PR reduces them from 189 weighted issues to 150. I started looking at the rest of the list but there's a combination of a high rate of false positives (various things like detekt wanting us to remove `suspend` from functions that absolutely need `suspect` as per "unable to compile without it", supposedly unreachable code that's behind early-return checks, things like that) as well as things that would require at least some amount of actual refactoring (threading injected dispatchers through various constructors, etc.). This PR is basically mechanical fixes with no change in logic. The one exception is that we had an UnsafeCallOnNullableType warning and I added a checkNotNull().

Alternatively, if we don't care about any of this or the remaining issues, we can close this PR and just remove the detektAndroidHostTest target completely.

### DIFF
--- a/composeApp/src/iosMain/kotlin/io/music_assistant/client/utils/PlatformDispatchers.ios.kt
+++ b/composeApp/src/iosMain/kotlin/io/music_assistant/client/utils/PlatformDispatchers.ios.kt
@@ -3,7 +3,6 @@ package io.music_assistant.client.utils
 import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Runnable
 import platform.darwin.DISPATCH_QUEUE_PRIORITY_HIGH
 import platform.darwin.dispatch_async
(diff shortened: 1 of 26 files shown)


composeApp/src/iosMain/kotlin/io/music_assistant/client/utils/PlatformDispatchers.ios.kt
```diff
@@ -3,7 +3,6 @@ package io.music_assistant.client.utils
 import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Runnable
```
formatBCE, line 6: Looks like iOS compilation wants this back?

teancom: @formatBCE *sigh* Yeah, fixed.

formatBCE (approved):

PR #747, opened 2026-07-06
Remove album/playlist sort menu

seadowg:
Based on discussion on Discord.

It seems like the simplest thing to do while keeping options open for later was just to remove all the sort orders except for "Original" and then not show sort chips when there's only one option. 


seadowg:
@formatBCE it seemed like you were up this, but no worries if it still feels controversial! One of those ones where I was just curious about how to do it, and then did it.

formatBCE (approved):
Okay, let's do it for now, and then re-introduce :)

PR #749, opened 2026-07-06
Don't reauthenticate with token when server ID has changed

seadowg:
This prevents weird (or potentially malicious) scenarios where a user joins a network with an MA server set up using the same local name as their own - not unlikely if people are using HA and have stuck with the default `homeassistant.local` address.

In that case, users will now get disconnected and shown an error. They can resolve this (if they do want to connect to the server) by deleting the connection and reconnecting. As well as the automated test, I've simulated a few different scenarios that this could come up in:

- Reinstalling: New server (ID) at same address with same username and password
    - Have to delete saved connection to access - no way to log out we just see server ID error when connecting
- Reinstalling and changing credentials: New server (ID) at same address with different username and password
    - Same as above
- Changing app user: Same server but delete user and create a new on with a different username and password
    - We get :"Authentication failed" and a button to "Reloading authentication providers". Clicking that gets us to a connected, but logged out state.

The implementation relies on saving a server ID when a server is authenticated and then verifiying this against a server before automatically sending a token for reauthentication. If the server IDs don't match, the app will disconnect.

The current solution does not prevemnt the user from getting in a situation where they attempt to login to a different server with the same local name if they explicitly log out before connecting to it - that scenario feels pretty unusual and I'd imagine is most likely when a user is explicitly trying to switch servers.

### DIFF
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -299,5 +299,5 @@
   <string name="toast_marked_played">Marked as played</string>
   <string name="toast_marked_unplayed">Marked as unplayed</string>
   <string name="toast_no_uri">Media item has no URI</string>
-  <string name="server_id_mismatch_error">Blah</string>
+  <string name="server_id_mismatch_error">This server's ID does not match the one provided when first logging in!</string>
 </resources>
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -299,5 +299,5 @@
   <string name="toast_marked_played">Marked as played</string>
   <string name="toast_marked_unplayed">Marked as unplayed</string>
   <string name="toast_no_uri">Media item has no URI</string>
-  <string name="server_id_mismatch_error">This server's ID does not match the one provided when first logging in!</string>
+  <string name="server_id_mismatch_error">Server ID was changed. Re-authentication required.</string>
 </resources>
--- a/composeApp/src/commonMain/composeResources/values/strings.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings.xml
@@ -299,4 +299,5 @@
   <string name="toast_marked_played">Marked as played</string>
   <string name="toast_marked_unplayed">Marked as unplayed</string>
   <string name="toast_no_uri">Media item has no URI</string>
+  <string name="server_id_mismatch_error">Blah</string>
 </resources>
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
@@ -19,6 +19,8 @@ import io.music_assistant.client.imageloader.ImageCacheInvalidator
 import io.music_assistant.client.settings.ConnectionHistoryEntry
 import io.music_assistant.client.settings.ConnectionType
 import io.music_assistant.client.settings.SettingsRepository
+import io.music_assistant.client.ui.compose.common.DisplayString
+import io.music_assistant.client.ui.compose.common.toDisplayString
 import io.music_assistant.client.utils.AuthProcessState
 import io.music_assistant.client.utils.ConnectionData
 import io.music_assistant.client.utils.DataConnectionState
@@ -371,6 +373,10 @@ class KtorServiceClient(
         logger.i { "Playback inactive (state=${stateLabel(_sessionState.value)})" }
     }
 
+    override fun forceDisconnect(reason: DisplayString) {
+        TODO()
+    }
+
     /**
      * Called when the app returns to the foreground.
      */
@@ -407,7 +413,7 @@ class KtorServiceClient(
         SessionState.Disconnected.NoServerData -> "Disconnected.NoServerData"
         SessionState.Disconnected.Backgrounded -> "Disconnected.Backgrounded"
         SessionState.Disconnected.ByUser -> "Disconnected.ByUser"
-        is SessionState.Disconnected.Error -> "Disconnected.Error(${state.reason?.message})"
+        is SessionState.Disconnected.Error -> "Disconnected.Error(${state.reason?.toString()})"
         SessionState.Connecting -> "Connecting"
     }
 
@@ -599,7 +605,7 @@ class KtorServiceClient(
                         }
 
                         is TransportState.Failed -> {
-                            _sessionState.update { SessionState.Disconnected.Error(transportState.error) }
+                            _sessionState.update { SessionState.Disconnected.Error(transportState.error.message?.toDisplayString()) }
                             logger.i { "Transport→Failed → Disconnected.Error: ${transportState.error.message}" }
                         }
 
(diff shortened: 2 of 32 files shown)


composeApp/src/commonMain/composeResources/values/strings.xml
```diff
@@ -299,4 +299,5 @@
   <string name="toast_marked_played">Marked as played</string>
   <string name="toast_marked_unplayed">Marked as unplayed</string>
   <string name="toast_no_uri">Media item has no URI</string>
+  <string name="server_id_mismatch_error">This server's ID does not match the one provided when first logging in!</string>
```
seadowg: This text is definitely up for debate! I just wanted to get something basic in there to start.

composeApp/src/commonMain/kotlin/io/music_assistant/client/auth/AuthenticationManager.kt
```diff
@@ -85,22 +86,24 @@ class AuthenticationManager(
...
                                             log.i { "AwaitingAuth(NotStarted) — no saved token for server" }
                                         } else {
                                             log.i { "AwaitingAuth(NotStarted) — auto-login with saved token" }
-                                            authorizeWithSavedToken(token)
+
+                                            val currentServerId =
+                                                state.connectionData.serverInfo?.serverId
+                                            val previousServerId =
+                                                settings.getIdForServer(serverIdentifier)
+                                            if (previousServerId == null || currentServerId == previousServerId) {
```
seadowg, line 100: Checking for `null` here means that we let existing installations continue to work without having to force logouts. The server ID will be populated the next time the app automatically reauthenticates - this means there is a very, very small possibility that the user could end up saving a "bad" server ID if they ran into one of the scenarios decribed before reauthenticating. I think that's allowable.

docs/DEV-ENVIRONMENT.md
```diff
@@ -32,4 +32,14 @@ The project currently supports the iOS and Android targets. Common code is held
 
 Tests for shared multiplatform code live in the `composeApp` module's `commonTest` source set. These can be run locally in the JVM for the Android target using `./gradlew :composeApp:testAndroidHostTest`.
 
-Tests for Compose UI code are in the `androidApp` module. This is because [multiplatform Compose testing is currently still experimental](https://kotlinlang.org/docs/multiplatform/compose-test.html) and tests written using the multiplatform approach cannot be easily run in a local JVM yet without the desktop target (which this project doesn't use). The Compose tests can be run with `./gradlew :androidApp:testDebug`.
\ No newline at end of file
+Tests for Compose UI code are in the `androidApp` module. This is because [multiplatform Compose testing is currently still experimental](https://kotlinlang.org/docs/multiplatform/compose-test.html) and tests written using the multiplatform approach cannot be easily run in a local JVM yet without the desktop target (which this project doesn't use). The Compose tests can be run with `./gradlew :androidApp:testDebug`.
+
+## Running MA Server locally
```
formatBCE, line 37: Do we actually need this?

composeApp/src/commonMain/composeResources/values/strings.xml
formatBCE: I guess something like "Server ID was changed. Re-authentication required." would work better?

composeApp/src/commonMain/kotlin/io/music_assistant/client/api/KtorServiceClient.kt
```diff
@@ -407,7 +412,7 @@ class KtorServiceClient(
         SessionState.Disconnected.NoServerData -> "Disconnected.NoServerData"
         SessionState.Disconnected.Backgrounded -> "Disconnected.Backgrounded"
         SessionState.Disconnected.ByUser -> "Disconnected.ByUser"
-        is SessionState.Disconnected.Error -> "Disconnected.Error(${state.reason?.message})"
+        is SessionState.Disconnected.Error -> "Disconnected.Error(${state.reason?.toString()})"
```
formatBCE: Hmm. this might expose class name to the user, instead of easy-to-understand info. Are you sure we wanna do that?

formatBCE: Maybe at least do `state.reason?.message ?: state.reason?.toString()`?

composeApp/src/commonMain/kotlin/io/music_assistant/client/auth/AuthenticationManager.kt
```diff
@@ -130,6 +133,14 @@ class AuthenticationManager(
                             state.user?.let { user ->
                                 log.i { "Authenticated" }
                                 _authState.value = AuthState.Authenticated(user)
+
+                                val serverIdentifier = settings.getServerIdentifier(state)
+
+                                val serverId = state.connectionData.serverInfo!!.serverId!!
```
formatBCE: I super hate `!!`. Please handle nullability.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/settings/SettingsScreen.kt
```diff
@@ -247,9 +250,6 @@ fun SettingsScreen(goHome: () -> Unit, exitApp: () -> Unit) {
                             )
                         }
                         // If user changed connection info, don't auto-retry - let them manually retry
-                    } else if (sessionState is SessionState.Connected) {
```
formatBCE, line 250: Hmm, why?

formatBCE (changes requested):
Couple comments.

docs/DEV-ENVIRONMENT.md
seadowg: I found it super useful to quickly create/teardown a server for simulating scenarios - thought it could be helpful for others. I can just keep it in `.local_gitignore` if you'd prefer it out of the project though.

composeApp/src/commonMain/kotlin/io/music_assistant/client/auth/AuthenticationManager.kt
seadowg: I'd meant to come back to this actually as it's a weird case. Do you think we can safely require `serverId` for `ServerInfo`? As far as I can see we always expect it.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/settings/SettingsScreen.kt
seadowg: Sorry I should have left a comment here! So, this can put us in a loop if the server disconnects during auth, but can successfully connect again - I ran into that with these changes and I think it's likley there will be other cases that could cause it. It's not quite clear to me the flow this auto connection is trying to provide. Just an inital connection reattempt if we get kicked to the settings screen due to an error? There's no test and `KtorServiceClient` handles reconnecting and auto connection. Do you think we still need it at all?

composeApp/src/commonMain/kotlin/io/music_assistant/client/auth/AuthenticationManager.kt
formatBCE: If we say that server ID is always required - we can safely expect it. But then field should be non-nullable.

seadowg: > But then field should be non-nullable.

Yeah sorry that's exactly what I meant: can we switch it to `String` safely? Sounds like yes!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/settings/SettingsScreen.kt
formatBCE: It was resetting reauth, if connected successfully. Don't really know if we need it anymore - let's keep an eye on it.

formatBCE (approved):

formatBCE:
Thanks for addressing everything!

PR #754, opened 2026-07-07
Fix screens not using full height

### LINKED ISSUE #753: Background not fully drawn during navigation animation on short-content views
## Context
 - Operating system and version: iOS/iPadOS 26.5.2
 - App version: TestFlight 1.0 (42)
 - Music Assistant server version: 2.9.5 Stable

### Problem description

Discovered on iPad, reproduced on iPhone with less content in a view. When navigating from, for example, `Home` to `Library`, during the animation, the background is only drawn for the part where there is content. When the animation is finished, the rest is filled in. This UI bug occurs on all screens that don't fill the entire visible view, noticed on navigating to `Home`, `Library`, and the `Search` page. It was most noticeable on `Library` on an iPad.
 
### Steps to reproduce

1. Make sure a view has only a single row of content, eg. hide everything but one item on the main `Library` view.
2. Navigate to another tab, `Home` or `Global Search`
3. Navigate back to the view with only a single row of content, in this example `Library`.
4. Notice how the background of the view you're navigating to 'misses' the bottom part, causing the previous view on the navigational stack to still be shown below it.

https://github.com/user-attachments/assets/eb0224f8-fe22-498f-ac2b-102baf275cc3

### Expected behavior

Ensure that the entire view has a drawn background already, so that the animation can be performed smoothly.

seadowg:
Closes #753

The nav "host" (`ConditionalBackNavDisplay`) wasn't using all the available space - this just fixes that.

Nope!

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
@@ -293,35 +293,32 @@ fun MainNavigationRoot(
                     playerExpanded = !playerExpanded
                 }
 
-                Box(
+                ConditionalBackNavDisplay(
                     modifier = Modifier
                         .fillMaxSize()
                         .background(MaterialTheme.colorScheme.background),
-                ) {
-                    ConditionalBackNavDisplay(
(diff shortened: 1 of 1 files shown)


formatBCE (approved):
Nice!

PR #760, opened 2026-07-09
Backfill login/logout tests and extract server settings from `ServiceClient`

seadowg:
Follow up to #749

The main idea here was to extract as much as possible of the `SettingsRepository` interactions from `KtorServiceClient` around connection/authentication to "thin out" those parts of the implementation. That simplifies that class itself, but more importantly moves more code within the boundary for tests that use `FakeServiceClient` (making those tests more "realistic").

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
@@ -16,7 +16,6 @@ interface ServiceClient {
     suspend fun authorize(token: String, isAutoLogin: Boolean = false)
     fun logout()
     val isReadyForCommands: StateFlow<Boolean>
-    val serverBaseUrl: StateFlow<String?>
     fun resolveImageUrl(
         path: String,
         provider: String,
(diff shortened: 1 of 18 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/api/ServiceClient.kt
```diff
@@ -40,6 +39,7 @@ interface ServiceClient {
     fun onExternalConsumerInactive()
     fun onPlaybackInactive()
     fun forceDisconnect(reason: Exception)
+    fun noServer()
```
seadowg, line 42: I'm not happy with the name of this method, but I'm having an absolute blank on a better one.

composeApp/src/commonMain/kotlin/io/music_assistant/client/connection/ConnectionManager.kt
```diff
@@ -0,0 +1,95 @@
...
+import io.music_assistant.client.webrtc.model.RemoteId
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+class ConnectionManager(
```
seadowg, line 18: This was designed to echo the approach taken by the `AuthenticationManager` to manage (or "coordinate") how we react to the changes in `SessionState` and persisting state about that domain in the `SettingsRespository`. 

I think more from `ServiceClient` could probably move in here: ViewModels could interact with this for connection logic instead of `ServiceClient`, it could have its own state abstraction for connection focused parts of the app to deal with (that don't need the full `SessionState` granularity). This feels like enough of a stab for now, but I can go further if it feels like that's needed before merging.

formatBCE (approved):

PR #771, opened 2026-07-13
Fix 'error loading data' against latest Music Assistant dev server

teancom:

The server (music-assistant/models dev) changed audiobook authors and narrators from list[str] to list[Artist | ItemMapping | str], so lists containing audiobooks (music/recommendations, audiobook library) failed to decode with 'Expected JsonPrimitive, but had JsonObject'.

Accept both shapes: plain strings pass through, objects contribute their 'name' field.


formatBCE (approved):
I hate when types of fields are getting changed. It's bad API architecture...
Thanks for handling this.

PR #773, opened 2026-07-13
Fix iPad crash when opening the log share sheet

teancom:

On iPad, UIActivityViewController is presented as a popover and UIKit throws an NSException in presentationTransitionWillBegin when no sourceView/sourceRect/barButtonItem anchor is set. Anchor the share sheet to the center of the root view with no arrow.


teancom:
Found this one digging through the test flight crash reports.

formatBCE (approved):

PR #774, opened 2026-07-14
Show audio processing details in mobile

marcelveldt:
> This draft PR is created for discussion and feedback on the audio processing data model.

Mobile playback details were inferred from partial queue data and could not describe shared processing or grouped player outputs.

- Consume versioned audio processing snapshots and updates
- Show input, shared processing, and per-player output details
- Keep the existing fallback for older servers


formatBCE (commented):
Sorry @marcelveldt  - it's huge job, but I will die on the hill that this HUGE change is bringing much more complexity, ambiguity and unnecessary data to the mobile app, than it brings profit to the user. For small dialog, that vast majority of users never gonna open, this PR adds 1.5k lines - extending and bringing hard-to-trace logic to the core classes like MainDataSource.

Are you 100% sure this functionality is this crucial?
Even if so - I would cut this down to:
- deserializing only absolutely necessary fields;
- moving the logic of stitching the data to the VM, instead of Main Data Source;
- refactoring/removing huge caching Store.

That's on the first glance. :)

@seadowg what do you say?

We can discuss in Discord, if it's easier. But I will fight. ;)

seadowg:
> @seadowg what do you say?

I do think we want a lot of this detail to allow people to inspect the audio chain without needing to switch to another client, but we hadn't yet discussed a design to incorporate it all (and hopefully combining it with DSP settings) - I was waiting for these server changes. On that, I'm happy to see the direction it's taken on the server! Looking at the interactions here, it looks like we're able to fetch the audio chain information through a command (by queue) and get an event every time it updates. It looks like we get easy access to everything in about inputs, outputs, server processing and DSP through that command/event data which is great.

I do agree with @formatBCE that this PR seems like more than we need right now though. We only display the quality in the expanded player, and so I think we can likely have a simpler approach where we just use the command to fetch audio chain info on demand, or use the event and "attach" information to our app level domain objects like `Player` or `Queue`. We could then bias towards a lot of the logic living at a display level (like in a ViewModel as @formatBCE suggests) and so we don't need to add another reactive data store like `AudioProcessingChainStore` really. 

I'd vote to close this right now rather than iterating on it, and then we can discuss how and what we want to display in mobile before anyone needs to do any work (and then I can write up some issues). We'll likely want to come back to this PR for at least as the `@Serializable` and API request structure if not more.

marcelveldt:
I've let my AI just create this PR to just settle on the data contract we discussed. Then I propose you only pick the parts you actually need/want to show

seadowg:
> I've let my AI just create this PR to just settle on the data contract we discussed. Then I propose you only pick the parts you actually need/want to show

Gotcha. I'll close for now, and we'll likely return to grab parts later. 

In future, it'd be helpful to declare if a PR is purely AI generated: there's an imbalance in effort if something is submitted without human oversight, but then humans (myself and @formatBCE) spend time reviewing it. Personally, knowing this PR was a generated "proof of concept" (which is what I'm interpreting from you here), would have changed how I'd reviewed and written about it.

marcelveldt:
The robot was a bit faster than I could write ;-)

formatBCE:
Yup, if I knew this PR is kind of documentation, I wouldn't read it line by line as I did - and my reaction would be different. :)

seadowg:
@marcelveldt looks like your robot is running wild again! Would be a good idea to make sure you've instructed it not to create PRs for client repos when working on server changes. If you're just proving things work for yourself, the agent should create PRs against your own fork (rather than upstream) for you to review there.

PR #779, opened 2026-07-15
iOS: route all nowPlayingInfo writes through NowPlayingInfoStore

teancom:

Introduce NowPlayingInfoStore as the single writer of MPNowPlayingInfoCenter.nowPlayingInfo. The store splits keys into two disjoint groups (track metadata vs transport), coalesces mutations into one assignment per main-loop pass, and stores position as a monotonic anchor (elapsed + timestamp + rate) projected at flush time so metadata writes can no longer rewind the progress bar. Transport rides the track lifecycle: it is rejected while no track is set and reset on track replacement. Main-thread use is enforced with dispatchPrecondition.

NowPlayingManager keeps all policy but no longer touches nowPlayingInfo directly. Adds an iosAppTests XCTest target with unit tests covering coalescing, anchor projection, clamping, and clear semantics.

### DIFF
--- a/iosApp/NowPlayingInfoStore.swift
+++ b/iosApp/NowPlayingInfoStore.swift
@@ -1,0 +1,126 @@
+import Foundation
+import MediaPlayer
+
+/// Sole writer of `MPNowPlayingInfoCenter.default().nowPlayingInfo`.
+///
+/// Two disjoint key groups: track metadata and transport (elapsed/rate). Position is
+/// held as an anchor and projected at flush; mutations coalesce into one dictionary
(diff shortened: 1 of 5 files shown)


formatBCE (approved):
Is it "store" in iOS? We always call it "repository" or "data source", depending on amount of the logic inside. :)
But I won't pretend that I know iOS naming patterns that good.

teancom:
Yup, *Store is idiomatic in iOS/Swift land for this sort of "object that owns access to a system-managed resource". EKEventStore, CNContactStore, HKHealthStore, StoreKit, etc.

PR #785, opened 2026-07-16
fix: improve reconnect resilience, HttpClient rotation, 20-attempt backoff, auto-resume

### LINKED ISSUE #693: Dropouts and connection issues - Help us understand your situation
## Goal of this issue

We read that some of our users struggle with dropouts or other connection issues. The first thing that often comes to mind is, ‘let’s increase the buffer size even further', which could help in some, but not all scenarios. To further investigate our options to improve the user experience, we would like to hear from you if you struggle with these kinds of problems.

### What we would like to know in case you experience similar issues

1. The platform you're on (Android/iOS).
2. What connection method you're using (Internal LAN, (reverse)proxy, VPN, etc.)
3. How you're connected (WiFi/Cellular).
4. What your average network speeds are when you experience such issues.
5. What codec you're using (Opus, FLAC, PCM).
6. Are there any noticeable patterns?

foobarth:
Closes #693 (at least remedies)

## Problem
3 compounding bugs prevent iOS MA Client from recovering after network drops (WireGuard tunnel drops, mobile network change, Wi-Fi roam, etc...):

1. **Stale HttpClient connection pool**: ktor/Darwin/NSURLSession caches DNS failures (Code=-1003). All subsequent reconnect attempts fail with same error even after the network is restored.
2. **Fixed 10-attempt ceiling**: ~3 minute window is too short for tunnel-based connectivity (5-15 min).
3. **No auto-resume**: Sendspin WS reconnects but doesn't know music was playing, so stream stays silent.

## Changes

* HttpClient rotation on network change + 20-attempt backoff + 500ms DNS grace
* Configurable reconnect with infinite retry mode (-1)
* Auto-resume playback after WS reconnect (Sendspin) within the first `RECONNECT_AUTO_RESUME_MAX_ATTEMPTS` attempts (currently 9)

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
@@ -178,6 +178,9 @@ class SendspinClient(
                             is SendspinState.Connecting,
                             is SendspinState.Reconnecting,
                             is SendspinState.Error -> {
+                                val wasStreaming =
+                                    (_state.value as? SendspinState.Reconnecting)?.wasStreaming
+                                        ?: false
                                 try {
                                     if (config.requiresAuth) {
                                         _state.update { SendspinState.Authenticating }
@@ -188,6 +191,15 @@ class SendspinClient(
                                     }
                                 } catch (e: Exception) {
                                     logger.w { "Failed to send auth/hello (transport closed during handshake): ${e.message}" }
+                                }
+                                // Auto-resume playback if we were streaming before the disconnect
+                                if (wasStreaming) {
+                                    try {
+                                        mediaPlayerController.resume()
+                                        logger.i { "Auto-resumed playback after reconnect" }
+                                    } catch (e: Exception) {
+                                        logger.w(e) { "Auto-resume failed" }
+                                    }
                                 }
                             }
                             else -> Unit
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
@@ -36,6 +36,15 @@ class SendspinClient(
     override val coroutineContext: CoroutineContext
         get() = Dispatchers.Default + supervisorJob
 
+    companion object {
+        /** Max reconnect attempt index that still triggers auto-resume.
+         *  Matches attempt 9 = ~4 minutes on the backoff ladder.
+         *  Beyond this the outage is considered too long for safe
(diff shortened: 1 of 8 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
```diff
@@ -187,6 +193,15 @@ class SendspinClient(
                                 } catch (e: Exception) {
                                     logger.w { "Failed to send auth/hello (transport closed during handshake): ${e.message}" }
                                 }
+                                // Auto-resume playback if we were streaming before the disconnect
+                                if (wasStreaming) {
+                                    try {
+                                        mediaPlayerController.resume()
```
formatBCE, line 211: I'm hesitant in this. Basically if there's multi-minute delay (e.g. server reboot) - we will suddenly start playing again after some extended silence - many things could happen by then...

formatBCE (approved):
Great resilience changes.
One open question. Otherwise - good to merge.

formatBCE:
@foobarth will you address the comment?

foobarth:
Hey, sure. I will be back tomorrow.

composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/SendspinClient.kt
foobarth: You have a point there, but given the user `wasStreaming` by the time of the disconnect, at least my natural assumption would be it continues where it left off. Should we expire the auto-resume after let's say 5 minutes? less?

foobarth: I went ahead and added a configurable attempts gate on top of the existing reconnect backoff. Seemed to be the least invasive method of a sensible timeout. Currently set to 9 attempts, which is roughly ~4 minutes after which it will no longer auto-resume.

Let's wait for the CI to come green.

formatBCE: Yeah user was streaming, but something happened - and if user didn't actively participate in fixing it (e.g. didn't restart the app, for example) - that might as well mean they, for example, fell asleep. And then after 4 minutes of silence it will be pretty jarring to hear that music starting again.
We can leave it for now - I just have feeling that we'll have to return to it. :)

foobarth: Understood. Quite hard to please every user on this. I'd say we leave it on 9 attempts for now. After all, a four minute sleep is not too deep to get upset by your metal playlist out of a sudden. ;-)

Issue #775, opened 2026-07-14
Search bars are inconsistent


seadowg:
The Library and Search screen search bars are inconsistent both visually and in how they behave (when the clear button appears etc). This can be confusing and also means we have two implementations of the same thing.

seadowg:
@formatBCE I've just noticed the (really nice) new filter UI in Library. Should I port that over to Search as part of this you think?

formatBCE:
If it's applicable to the search (parameters wise) - then it would be great. :)

Thanks ;)

seadowg:
> If it's applicable to the search (parameters wise) - then it would be great. :)

Yeah, I think just moving the existing filter options (media types and library only) into a similar bottom sheet would be really nice!

formatBCE:
Probably worth making it modular and reuse?

seadowg:
> Probably worth making it modular and reuse?

For sure!

PR #786, opened 2026-07-16
Make search inputs consistent

seadowg:
Work towards #775

We're now using the same component in both places which just meant adapting the `ViewModel`s slightly used in Library/Search so they work with the `SearchInputs` input/output structure. 

I've also made `SearchInput` use a (mostly) customized `TextField` as opposed to the Material search default one. This gives us way more control over shape and size and allowed me to use slightly smaller text and drop the height to make the whole thing look less "bulbous".

I've also backfilled some tests for things I didn't want to break (or did break) while making these changes.

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
@@ -43,6 +43,7 @@ import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.encodeToJsonElement
+import kotlin.collections.addAll
 
 class FakeServiceClient : ServiceClient {
     private var legacyVersion: LegacyVersion? = null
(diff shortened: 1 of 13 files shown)


androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
```diff
@@ -55,11 +56,18 @@ class FakeServiceClient : ServiceClient {
...
     private val queueItems = mutableMapOf<String, List<ServerQueueItem>>()
     private val items = mutableSetOf<ServerMediaItem>()
+    private val globalItems = mutableListOf<ServerMediaItem>()
+
     private val albums: List<ServerMediaItem>
         get() {
             return items.filter { it.mediaType == MediaType.ALBUM.serverValue }
         }
 
+    private val globalAlbums: List<ServerMediaItem>
```
seadowg, line 66: I'll expand this concept a little more when reworking the filter UI.

seadowg:
@formatBCE I haven't had a chance to look at the filter UI yet and I don't think I will for a couple more days at least so just sticking this in now as it already has nice improvements for the search inputs.

formatBCE (approved):
Good refactoring!

PR #788, opened 2026-07-17
 Move iOS now playing rendering onto the state channels

teancom:

This is the actual transition to using the new approach. I've been living-on testing it for two days and it seems to be solid.

Rename NowPlayingManager to NowPlayingCoordinator and hand each system-media surface to a channel handler: track metadata (with URL-cached, token-guarded artwork loads and anti-flicker deferral), transport anchors (re-stamped on arrival with the Swift clock, restored with forward projection after a rebuilding track write), and queue modes (sole writer of shuffle/repeat command state). Shuffle and repeat toggles round-trip through new toggle_shuffle/toggle_repeat commands mapped in LocalPlayerController via an extracted pure function. The legacy updateNowPlaying receivers become no-ops, the remote-seek direct dictionary write is gone, and lock screen seek now relies on the Kotlin freeze anchor.

Two more PRs left - one to delete all of the old code (done separately to keep this one reviewable) and one to actually add the CarPlay shuffle/random buttons.

More logging! Tracking down one bug led me to discover we had every 'entry' to commands covered except two, so figuring out where an errant playback came from was harder than it should have been.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -240,6 +240,8 @@ class HomeScreenViewModel(
         dataSource.selectedPlayer?.queueOrPlayerId?.let { queueId ->
             item.mediaUri?.let { mediaUri ->
                 viewModelScope.launch {
+                    Logger.withTag("PlayDispatch")
```
formatBCE, line 243: I see many logs with PlayDispatch. Are you intended to keep them, or they're part of debugging?

formatBCE (approved):
If logs are intended - feel free to merge.

PR #789, opened 2026-07-18
Tolerate missing legacy stream DSP

marcelveldt:
Queue stream details will stop including the legacy top-level DSP map.

- Accept payloads where `dsp` is omitted
- Preserve decoding of legacy per-player DSP payloads

### DIFF
--- a/composeApp/src/commonTest/kotlin/io/music_assistant/client/data/model/server/ServerQueueItemSerializationTest.kt
+++ b/composeApp/src/commonTest/kotlin/io/music_assistant/client/data/model/server/ServerQueueItemSerializationTest.kt
@@ -1,0 +1,65 @@
+package io.music_assistant.client.data.model.server
+
+import io.music_assistant.client.utils.myJson
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class ServerQueueItemSerializationTest {
+    @Test
+    fun deserializesStreamDetailsWithoutLegacyDsp() {
+        val json = """
+            {
+              "queue_item_id": "queue-item-1",
+              "streamdetails": {
+                "provider": "library",
+                "audio_format": {
+                  "content_type": "audio/flac",
+                  "sample_rate": 44100
+                }
+              }
+            }
+        """.trimIndent()
+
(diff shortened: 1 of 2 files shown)


formatBCE (approved):

composeApp/src/commonTest/kotlin/io/music_assistant/client/data/model/server/ServerQueueItemSerializationTest.kt
```diff
@@ -0,0 +1,65 @@
+package io.music_assistant.client.data.model.server
+
+import io.music_assistant.client.utils.myJson
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class ServerQueueItemSerializationTest {
```
seadowg, line 8: I think it'd be better to test this against `FakeServerClient` which has a mechanism already for dealing with "legacy" API response tolerance. This test only validates that the `ServerQueueItem` deals with the variance in API response whereas we want to make sure that's the case for the "full" (at least most of the) stack of the app and also make it clear that the test and functionality can be removed when we stop supporting that legacy API version.

seadowg (changes requested):

composeApp/src/commonTest/kotlin/io/music_assistant/client/data/model/server/ServerQueueItemSerializationTest.kt
seadowg: I'm happy to add to do the work to make that change and push it here (assuming we have that enabled for this repo) if that's helpful!

formatBCE: > I'm happy to add to do the work to make that change and push it here (assuming we have that enabled for this repo) if that's helpful!

We can always merge, and then build your PR on top. Or make changes here if you can. Functionally this is small change - feel free to do whatever you want ;)

seadowg: Yeah true! I'm happy to follow up.

seadowg (approved):

PR #802, opened 2026-07-21
Add tests for audio chain

seadowg:
Follow up to https://github.com/music-assistant/mobile-app/pull/789

I'd like to be able to have all the testing against `FakeServiceClient` here, but that results in us having to deal with maintaining different serialization classes (or handwriting JSON) for responses. I think the mix of behavioural tests using our "tolerant" serialization classes and a check for that the latter works with both API version responses is probably the right mix for now.

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/feature/AudioChainTest.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/feature/AudioChainTest.kt
@@ -5,6 +5,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import io.music_assistant.client.api.ServiceClient
 import io.music_assistant.client.data.model.server.AudioFormat
 import io.music_assistant.client.support.FakeServiceClient
+import io.music_assistant.client.support.FakeServiceClient.LegacyVersion
 import io.music_assistant.client.support.Qualifiers
 import io.music_assistant.client.support.ServerMediaItemFixtures
 import io.music_assistant.client.support.ServerPlayerFixtures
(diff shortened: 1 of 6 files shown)


androidApp/src/test/kotlin/io/music_assistant/client/feature/AudioChainTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    @get:Rule
+    val testRuleChain = createTestRuleChain()
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    private val serviceClient: FakeServiceClient by inject(ServiceClient::class.java)
+
+    @Test
+    fun `does not crash`() {
```
seadowg, line 33: We don't support the audio chain in 2.10 yet, so this just checks that nothing explodes when using the app with it. I'll write some follow-up issues for the new audio chain API integration soon.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerQueueItem.kt
```diff
@@ -34,7 +34,7 @@ data class StreamDetails(
 //    @SerialName("target_loudness") val targetLoudness: Double,
 //    @SerialName("strip_silence_begin") val stripSilenceBegin: Boolean,
 //    @SerialName("strip_silence_end") val stripSilenceEnd: Boolean,
-    @SerialName("dsp") val dsp: Map<String, DSPSettings> = emptyMap(),
+    @SerialName("dsp") val dsp: Map<String, DSPSettings>? = null,
```
seadowg, line 37: I think making this nullable makes it clearer to calling code that it's not available.

formatBCE (approved):

PR #809, opened 2026-07-23
Android TV / Google TV support

### LINKED ISSUE #808: Android TV / Google TV support
## Summary

The app currently has zero Android TV / Google TV awareness, which makes a sideloaded install
unreliable there today:

- No `LEANBACK_LAUNCHER` intent filter, so the app doesn't appear on the TV home screen's app row.
- The camera feature is hard-required in the manifest, which blocks the Play Store listing on
  TV devices (none of which have a camera).
- `MainActivity` force-locks portrait orientation on phone-width heuristics that don't apply to
  TV screens, which are always landscape.
- The UI is touch-only — nothing in the app claims D-pad/keyboard focus, so a TV remote's D-pad
  has nothing to move between.

## Context

- #689 ("Android TV") was closed not-planned.
- #405 ("Google TV support") was closed completed, but the current manifest/UI don't reflect
  that — no leanback entry, no TV-specific handling.
- #5162 — a maintainer confirmed the sideloaded app is unreliable on TV today, specifically
  calling out that the D-pad up/down on the now-playing screen moves the volume slider instead
  of navigating, because nothing else claims focus.

Given that history, it's worth re-opening this as a scoped, concrete piece of work rather than
a general "support Android TV" ask.

## Proposed scope

I have a branch with a first pass at this:

1. Manifest changes for TV eligibility: camera/touchscreen/leanback declared optional, added the
   `LEANBACK_LAUNCHER` intent filter and an `android:banner`.
2. `MainActivity` detects `Configuration.UI_MODE_TYPE_TELEVISION` and forces landscape instead
   of relying on width-based heuristics.
3. The WebRTC remote-pairing QR-scan button is hidden when there's no camera (`hasCamera()`
   expect/actual), since it's a dead end on TV — the manual remote-ID field stays available.
4. A fix for the D-pad-hijacks-the-volume-slider bug from #5162: the player-switcher dialog now
   claims initial D-pad focus when it opens.

Item 4 is only a partial fix — initial focus now lands correctly, but D-pad DOWN doesn't yet
move between rows in that dialog (still leaks focus to whatever's behind it). That's a
`ReorderableItem`/focus-traversal issue I haven't tracked down yet.

## Open question before going further

Is a player switcher the right thing to expose on TV at all? On phone/tablet it makes sense —
you're choosing among several Sonos-like rooms/players. A TV is inherently one fixed playback
target, so the dialog (and the focus-traversal bug in it) may not be worth polishing for this
form factor — it might be more correct to just not surface the switcher on TV, and spend the
D-pad-focus effort on the parts of the UI you actually use from a couch: home/library/search
navigation and playback controls.

Wanted to check in on that before sinking more time into a dialog that might not belong here.
Happy to open a PR with what's above (plus fixing the traversal bug, or removing the switcher
on TV — whichever direction makes sense) once there's agreement on approach. If it turns out
there is a real need for player switching on TV, I'd suggest tracking the D-pad traversal fix
as its own follow-up issue rather than folding it into this one.

dmfrey:
Draft PR for #808 — opening early for visibility while the open question there gets settled,
not because this is ready to merge.

## What's here

1. Manifest changes for TV eligibility: camera/touchscreen/leanback declared optional, added
   the `LEANBACK_LAUNCHER` intent filter and an `android:banner`. Includes a fix for a manifest
   merger issue where `ktor-client-webrtc`'s own implicit `required="true"` on the camera
   feature was silently overriding the app's `required="false"`.
2. `MainActivity` detects `Configuration.UI_MODE_TYPE_TELEVISION` and forces landscape instead
   of relying on phone-width heuristics.
3. The WebRTC remote-pairing QR-scan button is hidden when there's no camera — a dead end on
   TV. The manual remote-ID field stays available either way.
4. A partial fix for the D-pad-hijacks-the-volume-slider bug from #5162: the player-switcher
   dialog now claims initial D-pad focus when it opens.

All four have unit/Robolectric test coverage, and this has been run for real on a Google TV
emulator (not just Robolectric) — app launches via the leanback launcher, locks to landscape,
connects and logs into a real server, navigates via D-pad.

## What's not here / known issues

- **The player-switcher D-pad fix is incomplete.** Initial focus lands correctly, but D-pad
  DOWN doesn't move between rows in that dialog — it currently leaks focus to whatever's behind
  it instead. Likely a `ReorderableItem` (`sh.calvin.reorderable`) focus-traversal interaction
  I haven't tracked down yet.
- Marked draft pending the open question in #808: whether a player switcher belongs in the TV
  nav at all, given a TV is one fixed playback target rather than several rooms/players to
  choose between. If the answer is "no," item 4 above and the bug in it are probably moot and
  this PR should drop that piece rather than fix it further.
- Nav rail/bar focus order, playback control focus indication at 10-foot viewing distance, and
  search/library grid D-pad traversal are all out of scope here — none of them blocked basic TV
  usability the way manifest eligibility and the switcher bug did.

Closes #808 once the above is resolved one way or the other.


androidApp/src/main/kotlin/io/music_assistant/client/MainActivity.kt
```diff
@@ -111,6 +109,22 @@ class MainActivity : ComponentActivity() {
...
     }
 }
+
+/**
+ * TVs are always landscape and never rotate, so that takes priority over the phone-width
+ * check. Compact (phone-width) devices are locked to portrait. Larger/tablet configurations
+ * are left as the platform default (free rotation) — hence the nullable return.
+ */
+internal fun resolveOrientationLock(configuration: Configuration): Int? = when {
+    configuration.isTelevision() -> ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE
```
formatBCE, line 122: Are you sure everyone will be using TV in landscape?

formatBCE (commented):
Good code.

androidApp/src/main/kotlin/io/music_assistant/client/MainActivity.kt
dmfrey: No — landscape is actually mandatory for Android TV apps, not just a convention. Google's Android TV app quality guidelines require that no activity declare `portrait`, `reversePortrait`, `sensorPortrait`, `userPortrait`, or `reverseLandscape` — landscape-only is a Play Store TV listing requirement. Most TV box firmware forces landscape at the OS level regardless of what an app declares, too. So this isn't an assumption on my part, it's just implementing what's already required for this form factor.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/MainNavRoot.kt
```diff
@@ -202,8 +264,12 @@ fun MainNavigationRoot(
             DeepLinkDestination.Players -> {
                 // Expand the now-playing layout over the current tab (the
                 // FloatingBar is global, so no tab switch needed). The pager
-                // renders its own empty state if no player is present.
+                // renders its own empty state if no player is present. Open the
+                // queue too: this path is the "return to a playing app" entry
+                // (notification tap / TV now-playing card), where the queue
+                // alongside the transport is the useful landing.
                 playerExpanded = true
+                isQueueExpanded = true
```
formatBCE, line 275: this shouldn't matter for landscape layout - queue there is always shown.

formatBCE:
Sorry @dmfrey I think these changes really bring a lot of complexity. IDK if we will be able to support this.
@seadowg what do you think?
@marcelveldt do we want to merge this for TV support?

seadowg:
> I think these changes really bring a lot of complexity. IDK if we will be able to support this.

@formatBCE agreed. I've only had time to skim, but I do think these changes are too wide and not isolated enough to fall into the category of a small contribution (and with that a low amount of risk of later maintenance costs). I'd opt to close this for now, but @dmfrey is still able to maintain it for themselves as a fork of course!

dmfrey:
Thanks for the honest feedback, both — I get the concern, and I don't want to push a big surface-area change onto you to maintain long-term.

Looking back through the diff, I think a good chunk of the "wide" feeling is real but fixable: a few files (`SettingsScreen.kt` especially, and `AuthenticationPanel.kt`) ended up with TV-specific logic interleaved inline instead of cleanly forked out, so the diff touches shared code paths more than it needs to. Some sections already fork cleanly (e.g. `DirectConnectionContent`/`DirectConnectionContentTv`) — I just didn't finish applying that pattern everywhere.

I'm moving this back to draft while I take a pass at minimizing the touchpoints. Plan is to pull the TV-only branches out of the shared composables into sibling `Tv*` files/functions with an early-return gate, so the phone/iOS code paths in those files go back to being untouched. That should shrink the diff in the two biggest offenders substantially without changing behavior. I'll ping this thread and mark it ready for review again once it's done.

No worries either way if you'd rather close it in the meantime — just didn't want to leave it hanging without a plan.

formatBCE:
@dmfrey you will be working on this anyway - so let's keep it drafted, and check after. :)

dmfrey:
Follow-up on the slimdown, plus some real fixes that came out of it.

## Diff-size cleanup (what I said I'd do)

- `MainNavRoot.kt` / `SelectPlayerDialog.kt`: the two TV-only focus effects/D-pad interception now gate behind `isTelevisionDevice()` instead of running unconditionally on every platform.
- `SettingsScreen.kt`: pulled the ~115-line focus-adjacency-map builders into `SettingsTvFocusLinks.kt` (pure functions), and moved the already-forked `*Tv` composables into `SettingsScreenTv.kt`. Net: `SettingsScreen.kt` goes from 2205 → 1604 lines (-27%).
- Went back through `LoginSection`/`MiscSection`/`ConnectionMethodTabs` looking for more to fork out and concluded they don't need it — `LoginSection`/`MiscSection` already use a safe nullable-flow pattern that's a true no-op on phone/iOS, and `ConnectionMethodTabs` has exactly one TV-specific line out of ~140. Didn't want to churn the diff further just for the sake of it.

## Real bugs found by actually using the app on hardware, not just reading the diff

Ran hands-on usability passes on both a Google TV emulator and a real Chromecast with Google TV — screenshots, `uiautomator` focus-state ground truth, functional select-tests — and found three real D-pad navigation bugs the original PR hadn't surfaced:

1. The persistent mini-player was **entirely unreachable via D-pad** from Home, Library, or Settings — only reachable by touch. Wired explicit DOWN/UP links between the nav rail, the last content row, and the floating bar.
2. Several elements (content cards, nav rail icons, the Home top-bar buttons) had **no visible focus indicator at all** — D-pad navigation technically worked but looked exactly like a dead remote. Gave `tvFocusRing()` an opt-in "track descendant focus" mode so it can be applied once at a couple of shared wrapper composables instead of duplicated into every per-media-type card.
3. The nav rail's own internal UP/DOWN traversal (Home/Library/Search/Settings) had never been given explicit directional links, unlike every other screen here — it relied solely on the same geometric-search-is-unreliable-on-this-hardware issue this PR already works around elsewhere. Pressing UP from the last item didn't just fail to move focus, it dropped it entirely. Added the same explicit-link pattern used throughout the rest of the app.

All three verified directly on the real device (not just the emulator) with before/after screenshots and functional tests, not just code review.

One known issue remains open and unresolved: a focus ring occasionally stays visually stuck for a moment exactly when D-pad DOWN crosses from the last content row into the floating bar (reachability itself works, confirmed functionally). Investigated it, including a fix attempt that didn't hold up under re-testing so I reverted it rather than ship something unverified. Minor/cosmetic, not a blocker.

Everything's verified with `detekt` + the full unit test suite, and CI is green. This should be ready for another look whenever you have time — taking it out of draft now.

formatBCE:
@dmfrey I'm sorry you have to deal with constant changes - we're moving fast, and breaking stuff even faster... :)

Do you want to continue on this?

dmfrey:
I've been keeping up with the changes and make sure to incorporate them. I'll keep it going.

Issue #801, opened 2026-07-21
Show all albums from all providers for an artist


seadowg:
## Problem

Currently, when clicking on an artist, we see library, all and top albums/tracks. There are three problems here that occur when an artist is matched across or within providers:

1. All will just show the "All" items from one of the providers. The same is true for "Top".
2. If an artist has been matched within a provider (two artist IDs on the same provider), "All" (and "Top") will not show the items from one of the artist IDs.
3. It's not clear where items are coming from in terms of provider.

## Solution

- [x] There should be just two sections for albums: "In your library" and "All albums". The "All albums" section should have a subtitle with the provider (like "On Spotify") and show the icon for the provider. There should be a control to allow the user to switch which provider is used for "All albums".
- [x] Tracks have the same rules as above except for just using "Top" from providers instead of "All".
- [x] If an artist has multiple IDs in the same provider (multiple provider mappings with the same provider ID), all the albums should be shown in the section for the provider.

We might want to experiment with using tabs instead of sections (so we'd have something like "Library", "All albums" and "Top tracks" tabs).

Bonus feature: if the set of items for one section exactly matches the set for "In your library" (the provider mappings exactly match), don't show that section. This prevents duplication when opening an artist that where every album from a provider is in your library. This is a problem on the web currently as well.

seadowg:
@formatBCE can you think of edge cases I'm missing here or anything that feels wrong to you?

formatBCE:
I don't think this is like it works on web. The logic now is pretty close.

seadowg:
> I don't think this is like it works on web. The logic now is pretty close.

Hmmm yeah honestly it does look like web is also showing "All albums" from just one source even when there are matches across providers - I don't think that's good though. Plus, I don't think we need to match the exact behaviour as we have a different layout for artists anyway.

I've got a concrete example here to demonstrate the problem. Luckily, I'm currently trying out Qobuz as a replacement for Tidal, so I've been able to set up both providers alongside OpenSubsonic (Navidrome) and done "Search all providers for matches" for Deftones:

<img width="1709" height="989" alt="Image" src="https://github.com/user-attachments/assets/dd2e666d-7116-44f0-8f74-bace47519d8e" />

Let's just skip over the fact that for some reason web is only showing albums for OpenSubsonic and I don't get access to **any** of the albums in Qobuz/Tidal - that's another issue I'll need to file elsewhere.

[...]

PR #816, opened 2026-07-24
Backfill artist albums tests

seadowg:
Work towards #801

The big change here to get these tests in a good place was to improve `FakeServiceClient` so that it handles the MA server concepts of the library and providers more like the real thing. With that, comes the ability to write tests that make better assertions about media items (identifying multiple on a screen, checking their provider domain etc).

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeMediaItemStore.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeMediaItemStore.kt
@@ -1,0 +1,156 @@
+package io.music_assistant.client.support
+
+import io.music_assistant.client.data.model.client.MediaType
+import io.music_assistant.client.data.model.server.ServerMediaItem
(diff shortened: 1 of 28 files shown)


androidApp/src/test/kotlin/io/music_assistant/client/support/FakeMediaItemStore.kt
```diff
@@ -0,0 +1,156 @@
+package io.music_assistant.client.support
+
+import io.music_assistant.client.data.model.client.MediaType
+import io.music_assistant.client.data.model.server.ServerMediaItem
+import io.music_assistant.client.utils.UniqueIdGenerator
+
+internal class FakeMediaItemStore {
```
seadowg, line 7: Pulling this out should allow `FakeServiceClient` to focus on request/response faking and not have to overly concern itself with the data model that we need to fake in tests.

formatBCE (approved):

PR #830, opened 2026-07-28
Close app on incompatible schema

formatBCE:


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/SchemaVersionWarningViewModel.kt
```diff
@@ -3,28 +3,47 @@ package io.music_assistant.client.ui
...
-            .map { schema -> schema?.takeIf { it > MAX_SUPPORTED_SCHEMA_VERSION } }
             .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), null)
+
+    /** Terminal check first: below the server's floor supersedes merely trailing its current schema. */
+    private fun classify(info: ServerInfo): SchemaWarning? {
+        info.minSupportedSchemaVersion?.let {
+            if (LOCAL_SCHEMA_VERSION < it) return SchemaWarning.CLIENT_INCOMPATIBLE
+        }
+        info.schemaVersion?.let {
+            if (LOCAL_SCHEMA_VERSION < it) return SchemaWarning.SERVER_AHEAD
```
seadowg, line 45: It sounds like from our discussions that we don't expect this case to be problematic (as long as we have our `@Serializable` classes correct) as schema version increments won't introduce backwards incompatible changes. Are you just worried about new features not being available here?

formatBCE: I added this before our conversation. I think this might be over cautious... I will delete this check.

PR #833, opened 2026-07-29
fix: send available field in client/state per spec

foobarth:
Add the required 'available' boolean to ClientStatePayload and pass it as true when sending client/state. This brings the mobile app in line with the sendspin spec, eliminating non-compliant client warnings from the server.

Backwards compatible: old servers that don't have the field ignore the extra JSON key; new servers use it as the primary availability signal.

Forward compat with sendspin spec.

No


formatBCE (approved):
Nice, thanks!

PR #834, opened 2026-07-29
Show "Discography" and "Top tracks"  for all providers

### LINKED ISSUE #801: Show all albums from all providers for an artist

seadowg:
Closes #801

TODO:

- [x] Add provider switching for top tracks
- [x] Add "View all" link for sections
- [x] Limit section items
- [x] Flesh out item list screen
- [x] Handle empty sections and empty provider lists
- [x] Show first non-empty provider for "Discography" and "Top tracks"

Follow-ups:
- [x] Add sort/view mode to list screen
- [x] Add tests for `ItemUseCases`
- [ ] Show provider names (rather than domain) in filter
- [x] Use reactive version of `MediaItemRepository#fetchMediaItems` everywhere

- I've expanded our `CategoryRow` composable to be able to handle `DataState` for loading, empty `NoData` states
- There's a new shared composable `ItemList` that is used by the "View all" screens for the artist as well as the library screens

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerMediaItem.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerMediaItem.kt
@@ -107,9 +107,7 @@ data class ServerMediaItemImage(
 @Serializable
 data class ProviderMapping(
     @SerialName("item_id") val itemId: String,
-    // Required by the server when a media_item is sent back (e.g. music/mark_played),
-    // so it must be retained on parse to round-trip in provider_mappings.
-    @SerialName("provider_domain") val providerDomain: String? = null,
+    @SerialName("provider_domain") val providerDomain: String,
     @SerialName("provider_instance") val providerInstance: String,
 //    @SerialName("available") val available: Boolean,
 //    @SerialName("audio_format") val audioFormat: AudioFormat? = null,
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.flow.scan
 import kotlinx.coroutines.flow.shareIn
 
 /**
@@ -47,6 +48,18 @@ class MediaItemRepository(
                 ?.let(factory::createList)
                 ?: error("Missing or undecodable media list payload")
         }
+
+    suspend fun fetchMediaItems(request: Request, observer: (List<AppMediaItem>) -> Unit) {
+        val result = fetchMediaItems(request)
+        val items = result.getOrNull()
+
+        if (items != null) {
+            observer(items)
+            itemChanges
+                .scan(items) { items, change -> items.replacing(change.item) }
+                .collect { observer(it) }
+        }
+    }
 
     /**
      * Issue [request] and decode its payload as a single client media item.
@@ -117,3 +130,6 @@ class MediaItemRepository(
         )
     }
 }
+
+private fun <T : AppMediaItem> List<T>.replacing(changed: T): List<T> =
+    map { if (it.itemId == changed.itemId) changed else it }
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/CategoryRow.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/CategoryRow.kt
@@ -1,5 +1,11 @@
 package io.music_assistant.client.ui.compose.common.items
 
+import androidx.compose.animation.animateColor
+import androidx.compose.animation.core.RepeatMode
+import androidx.compose.animation.core.infiniteRepeatable
+import androidx.compose.animation.core.rememberInfiniteTransition
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -7,13 +13,18 @@ import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.DropdownMenu
 import androidx.compose.material3.DropdownMenuItem
 import androidx.compose.material3.FilterChip
+import androidx.compose.material3.LocalTextStyle
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.ProvideTextStyle
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
@@ -23,11 +34,13 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.semantics
-import androidx.compose.ui.text.font.FontWeight
-import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import io.music_assistant.client.data.model.client.items.Album
 import io.music_assistant.client.data.model.client.items.AppMediaItem
@@ -41,28 +54,12 @@ import io.music_assistant.client.data.model.client.items.RadioStation
 import io.music_assistant.client.data.model.client.items.Track
 import io.music_assistant.client.ui.compose.common.DataState
 import io.music_assistant.client.ui.compose.common.DisplayString
+import io.music_assistant.client.ui.compose.common.toDisplayString
 import io.music_assistant.client.ui.compose.item.ItemList
 import musicassistantclient.composeapp.generated.resources.Res
 import musicassistantclient.composeapp.generated.resources.cd_view_all
 import org.jetbrains.compose.resources.StringResource
 import org.jetbrains.compose.resources.stringResource
-
-data class ItemCategory<T>(
-    val id: String,
-    val title: DisplayString,
-    val items: List<AppMediaItem>,
-    val list: ItemList? = null,
-    val filter: Filter<T>? = null,
-    val lazyListKey: String = id,
-    val tag: String? = null,
-) {
-    data class Filter<T>(
-        val label: DisplayString,
-        val options: List<T>,
-        val labelTransform: (T) -> DisplayString,
-        val contentDescription: StringResource,
-    )
-}
 
 @Composable
 fun <T, U> CategoryRow(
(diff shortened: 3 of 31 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/server/ServerMediaItem.kt
```diff
@@ -107,9 +107,7 @@ data class ServerMediaItemImage(
 @Serializable
 data class ProviderMapping(
     @SerialName("item_id") val itemId: String,
-    // Required by the server when a media_item is sent back (e.g. music/mark_played),
-    // so it must be retained on parse to round-trip in provider_mappings.
-    @SerialName("provider_domain") val providerDomain: String? = null,
+    @SerialName("provider_domain") val providerDomain: String,
```
seadowg, line 110: This is required as far as I can see

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -48,6 +49,18 @@ class MediaItemRepository(
                 ?: error("Missing or undecodable media list payload")
         }
 
+    suspend fun fetchMediaItems(request: Request, observer: (List<AppMediaItem>) -> Unit) {
```
seadowg, line 52: I think we can look at using this pretty much everywhere to avoid having to wire up specific change observations in every ViewModel that deals with item lists. This PR is already big enough though, so that can be a follow-up!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/CategoryRow.kt
```diff
@@ -27,28 +52,116 @@ import io.music_assistant.client.data.model.client.items.Podcast
...
 
-data class ItemCategory(
-    val id: String,
-    val title: DisplayString,
-    val items: List<AppMediaItem>,
-    val lazyListKey: String,
-    val tag: String? = null,
-)
+@Composable
+fun <T, U> CategoryRow(
```
seadowg, line 65: The thinking around the design here is that we can use this `DataState` overload for the artist screen (as I'm doing), but also for the home screen recommendations in future. In the latter case, my guess is that we'll likely end up with the ViewModel exposing a list of `DataState<RecommendationFolder>` (as opposed to `DataState<List<RecommendationFolder>>`) and that this pattern should work well for displaying that in a `LazyColumn`.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/items/CategoryRow.kt
```diff
@@ -27,28 +52,116 @@ import io.music_assistant.client.data.model.client.items.Podcast
...
+            onNavigateToList = onNavigateToList,
+            onOptionSelected = onOptionSelected,
+            onPlayClick = onPlayClick,
+            playlistActions = playlistActions,
+            libraryActions = libraryActions,
+            progressActions = progressActions,
+            providerIconFetcher = providerIconFetcher,
+        )
+    } else if (data is DataState.Loading) {
+        val placeholderWidth = 140.dp
```
seadowg, line 90: This placeholder is very basic - we can definitely iterate on making it fancier, but I think it's good enough for a first pass.

seadowg, line 90: I actually couldn't work out where the width/height for our grid items (albums, artists and so on) is coming from so the dimensions here are just eyeballed. It looks like we potentially just end up with the size we do because we fix a 1:1 aspect ratio and pull down a thumb sized image? If that's the case, we should make them fixed size so that we can use the same size for placeholders.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsScreen.kt
```diff
@@ -439,10 +385,30 @@ private fun ItemContent(
             // selectedTab is null exactly while sub-lists load (for an item that has tabs), so it
             // doubles as the loading gate: show the hero + a single spinner, tabs hidden.
             val currentTab = state.selectedTab
-            if (tabs.isEmpty()) {
+
+            val gridState = rememberLazyGridState()
+            if (item is Artist) {
+                ProvideClickActions(ClickContext.ARTIST) {
+                    ArtistContent(
```
seadowg, line 392: It honestly might be better to just pull this out as its own "screen", but I'm not sure enough that that's worth it yet.

formatBCE: I agree, it's so different now that having it here seems controversial

seadowg:
@formatBCE if you like, I think we could look at reviewing this as-is and I can follow up on the remaining todos. The biggest user facing thing that would be missing in that case would be sort/view grid controls for the "View all" screens.

formatBCE:
I installed it on device, and don't feel like I see everything...
When I look at artist in library, I only see library albums. That's it...
Shouldn't I see everything? It looks like it provides less info than it was there (at least I had tracks too, now it's just row of albums "in library"...

seadowg:
> I installed it on device, and don't feel like I see everything...
When I look at artist in library, I only see library albums. That's it...
Shouldn't I see everything? It looks like it provides less info than it was there (at least I had tracks too, now it's just row of albums "in library"...

Can you post a before and after screenshot or a list of the items you were seeing, and what isn't there after the changes? I don't quite understand the exact difference you're seeing here from this description.

formatBCE:
Same artist. Before: in lib, plus all:




After:

formatBCE:
Well, tracks invisible at all

seadowg:
@formatBCE Oh wow yeah that doesn't look right! That's really strange. It's like the artist doesn't have `providerMappings` or something. I've also got the Tidal provider, but I can't reproduce that with a-ha and I haven't seen it with any other artist either - for all artists I see "In library" (all the albums in my MA library), "Discography" (all albums on a provider) and "Top tracks" (the top tracks from a provider). Some questions:

- Is just a-ha like that or is every artist?
- If it's every artist, is that true for artists not in the library as well?
- Are you just using the Tidal provider or do you have other ones connected?
- Could you grab the API response for a-ha (or any artist if they are all affected) so we can see if that gives us any clues?

seadowg:
@remon1496 let me know if you've got some time to test this out as well and I'll send you an APK if so!

formatBCE:
- every artist
- if artist doesn't have library items, all are shown. Otherwise - only library albums (even if tracks are added explicitly).
- I have fair amount of local music
- will try to fix it

seadowg:
@formatBCE ah I'm wondering if you're getting zero items back for the first provider mapping and then the category is being hidden? That'd be a bug in how the code handles "empty" states, so I've removed that feature for now, and maybe you'll start seeing empty sections and be able to select the correct provider.

seadowg:
@formatBCE OK assuming that was the problem you were running into, I've now made it so that categories get hidden when empty only if they don't have a filter (like "In library"). We'll probably want to either filter out empty providers or add a nicer looking empty state (I think I prefer the latter to make sure it's clear to users what's happening), but this should mean you see all the items again. Let me know if that's looking better now!

formatBCE:
Should this work on current 2.9.10? Or only 2.10?

seadowg:
> Should this work on current 2.9.10? Or only 2.10?

I've only been testing with stable (2.9.10).

formatBCE:
Okay, now I see providers, and can open Tidal.
Did you just take first provider always? I guess we have to know counts for all providers (or actually load data for all providers right away in background).
Then we could filter out empty providers, and hide category only if all providers are filtered out.

seadowg:
> Did you just take first provider always? I guess we have to know counts for all providers (or actually load data for all providers right away in background).

Yeah, for the moment it just always starts with the first provider as otherwise we have to potentially hit multiple providers which felt slower as a starting point.

> Then we could filter out empty providers, and hide category only if all providers are filtered out.

I'm actually wondering if the next step might be to keep all providers, but persist the choice the user makes between screens so that. It might be worth seeing what feedback we get on this though. This will be very different for people depending on how many providers they use, if artists are matched across them, if providers are synced to the library and so on.

formatBCE:
Well, persisting provider should be done per-artist then... :(

Because if I have one artist in local lib, and other in Tidal, I will be constantly switching between, and annoyed.

seadowg:
> Because if I have one artist in local lib, and other in Tidal, I will be constantly switching between, and annoyed.

Yup! For some setups people will want consistency. Others not. That's why I'm thinking we wait to get feedback on how people use this.

formatBCE:
It's safer to show all from scratch.
We can show first provider, fetch data for it - then if it's empty, switch to next provider and remove empty one from list. It shouldn't be that bad - especially if we introduce not just items loading state, but row loading state too.

seadowg:
> It's safer to show all from scratch.

Agreed! I'll leave that as-is.

> We can show first provider, fetch data for it - then if it's empty, switch to next provider and remove empty one from list. It shouldn't be that bad - especially if we introduce not just items loading state, but row loading state too.

OK that sounds good - I'll make that change. The rows are already loading in separately with their own loading state so you're right that that should be fine even if it takes a little longer.

formatBCE:
I also have annoying bug: I find artist in global search, add album to the library, add track to the library. Then I go to library - artists, and search artist - and don't find it. 

It's strange and bummer.

formatBCE:
Oh, and we lost sorting, right? Now we're closer to the server view - but they have sorting for every row separately.
This grows so complicated. IDK is starts to be unpleasant to me to use the view.

seadowg:
> I also have annoying bug: I find artist in global search, add album to the library, add track to the library. Then I go to library - artists, and search artist - and don't find it.

Sounds like that must be server side right? We're hitting the server for search as far as I know so there shouldn't be anything state app side (unless it's http caching).

> Oh, and we lost sorting, right? Now we're closer to the server view - but they have sorting for every row separately.
This grows so complicated. IDK is starts to be unpleasant to me to use the view.

I'm thinking we add sorting into the screens linked with "View all" (I've got a todo for that in the PR description - I was asking whether we'd need that before merging) and just use the server response order or a default sort order for the rows. I agree that adding more controls just makes things too complicated! I want to keep the artist screen as an "overview" really, and you can drill down into the section you're interested in.

seadowg:
@formatBCE ok it now shows the first non-empty (or just the first if all are empty) provider for "Discography"/"Top tracks". Let me know how that's feeling. 

I've just spiked the change without tests - there's some bootstrapping to do to be able to write the granularity of test I'd like, and I think it's worth us being happy about the behaviour before I start put that effort in! That also means the code isn't as nice as I'd like it because I don't want to play with refactors without a safety harness.

formatBCE:
Yeah it looks better. 

On unrelated:
_I still have so many questions on the layout... I feel like we're going deeper into the navigation - now to see tracks by the actor I will have to go another click. And basically tracks for album and top tracks for artist, being both list of tracks, practically are different views - with their own view modes etc.
BTW, will there be favourites filter too? I tried to use this layout, and missed it right away. :)_

seadowg:
> I still have so many questions on the layout... I feel like we're going deeper into the navigation - now to see tracks by the actor I will have to go another click. 

Yeah, I think that's fair. My thinking is that this does mean adding another screen in many cases, but that navigating is more organized - getting to the list of things I want to be looking through should be easier/faster for artists with a lot of content. Right now we can end up with some very long lists of albums or tracks for artists with big catalogues that are hard to navigate.

I'm also wondering if you're feeling there should be another section for "Tracks in library" or just "Tracks" (as opposed to top tracks)? We could also just flip the order so "Top tracks" is first to make them more prominent.

> And basically tracks for album and top tracks for artist, being both list of tracks, practically are different views - with their own view modes etc.

Oh, yeah I think we want view modes for the "View all" pages. View mode toggles, sort and favourites (don't worry!) definitely, and then I'm thinking probably search and maybe other filters as well. Do you think we need any of that before merging? I've got a list of things in the PR description as "follow-ups", but let me know if you think they're blockers for merging this.

formatBCE:
Makes sense.
I think follow ups are ok - if they're fast, because we basically broke functionality without replacement.

But we need this pr merged so we could make that Home page working on beta - it's annoying.

seadowg:
@formatBCE ok yeah I think best to get it in then. I'll mark this ready for review. I can hopefully get some time to add sort/view mode on Wednesday. Either way it's the next thing I'll do.

formatBCE (approved):

formatBCE:
Good job. Let's merge :)

remon1496:
When only using a local file system provider, empty sections are visible. Shouldn't these be hidden if there is no content whatsoever?

image

seadowg:
@remon1496 is there content if you switch providers for "Discography" or "Top tracks"?

remon1496:
> @remon1496 is there content if you switch providers for "Discography" or "Top tracks"?

This instance only has a local file system provider. So in this situation I can't switch providers.

seadowg:
@remon1496 ah ok. Yeah, right now we never hide provider options as we were worried it might be confusing in cases where the user doesn't know if they have data available or not. We'll likely add a better looking empty state though.

remon1496:
> @remon1496 ah ok. Yeah, right now we never hide provider options as we were worried it might be confusing in cases where the user doesn't know if they have data available or not. We'll likely add a better looking empty state though.

Makes sense! So, not a bug. Good to know. :)

PR #835, opened 2026-07-29
Fix WebRTC message reassembly

teancom:

Reassemble chunked MA API frames before dispatch while preserving legacy whole-message handling. Bound incomplete groups and payload sizes, with coverage for malformed and interleaved chunks. Tested against both head of `dev` and on the stable 2.9.9 release.

N/A


formatBCE (commented):
I hate the fact that we must recreate WebRTC instead of using some library..

Good job, thank you!

formatBCE (approved):

PR #836, opened 2026-07-30
Fix iOS FLAC decoding crashes

teancom:

Serialize access to the stateful native FLAC decoder to prevent concurrent mutation of its pending data. Wait for the previous playback consumer to fully stop before resetting or replacing the decoder.

Got a few crash reports through iOS Test Flight, all from long-running (like, overnight) sessions, all from concurrent FLAC data mutation.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/AudioStreamManager.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/AudioStreamManager.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.delay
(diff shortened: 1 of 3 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/AudioStreamManager.kt
```diff
@@ -179,8 +180,11 @@ class AudioStreamManager(
         }
 
         try {
+            isStreaming = false
```
formatBCE: Makes sense extracting these three lines (and other occurence below) into function.

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/audio/AudioStreamManager.kt
teancom: I did it, but of the three places we stop playback in this file only two can use the function because the third needs cancel() not cancelAndJoin(). If you don't want a function for just two uses, I can remove the commit before merging.

formatBCE: 2 == duplication anyways.
I would just add boolean parameter "join: Boolean = true", and use "false" in red place. :)

teancom: Ah, got it. Done (and the previous Android CI failure was a flaky test that I'll look into and have a different PR for).

formatBCE (approved):

PR #843, opened 2026-08-05
Add view mode toggle to artist "View all" screens

seadowg:
Work towards #801
Follow up for #834 (more on the way)

I've extracted the repeated view mode `ViewModel` logic to its own `ViewModel` as well as creating a specific component for the toggle to make it as easy as possible to add this to new/existing screens when needed.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ViewModeViewModel.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ViewModeViewModel.kt
@@ -1,11 +1,12 @@
 package io.music_assistant.client.ui.compose.item
 
+import androidx.lifecycle.ViewModel
 import io.music_assistant.client.data.model.client.MediaType
 import io.music_assistant.client.settings.SettingsRepository
 import io.music_assistant.client.settings.ViewMode
 import kotlinx.coroutines.flow.StateFlow
 
-class ViewModeMediator(private val settingsRepository: SettingsRepository) {
+class ViewModeViewModel(private val settingsRepository: SettingsRepository) : ViewModel() {
     fun viewModeFor(mediaType: MediaType): StateFlow<ViewMode> {
         return settingsRepository.viewMode(mediaType)
     }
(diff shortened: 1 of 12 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ViewModeViewModel.kt
```diff
@@ -0,0 +1,18 @@
+package io.music_assistant.client.ui.compose.item
+
+import androidx.lifecycle.ViewModel
+import io.music_assistant.client.data.model.client.MediaType
+import io.music_assistant.client.settings.SettingsRepository
+import io.music_assistant.client.settings.ViewMode
+import kotlinx.coroutines.flow.StateFlow
+
+class ViewModeViewModel(private val settingsRepository: SettingsRepository) : ViewModel() {
```
seadowg, line 9: This is an absolute nightmare to type, but I can't think of a better name!

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/nav/TwoRowTopAppBar.kt
```diff
@@ -22,6 +23,7 @@ fun TwoRowTopAppBar(
     navigationIcon: @Composable () -> Unit = {},
     actions: @Composable RowScope.() -> Unit = {},
     secondRow: @Composable RowScope.() -> Unit = {},
+    secondRowArrangement: Arrangement.Horizontal = Arrangement.End,
```
seadowg, line 26: Right now we're always doing `Arrangement.End` so this saves us some repetition while also making it possible to easily tweak in the future.

formatBCE (approved):
You know, I didn't usually use separate VMs for extracted functionality before - and in this project it proves to be really useful. Don't remember where it started (probably with item actions), but it's good pattern. :)

seadowg:
> Don't remember where it started (probably with item actions), but it's good pattern. :)

Yeah I've found is super useful in the past to detach them from the actual screen domain bounds and have ones shared between screens, used on different screens and sometimes have multiple smaller ones for different small sections of screens. Otherwise, they can get real big, real fast for some UIs.

PR #844, opened 2026-08-06
Fix home page on Music Assistant Server 2.10 Beta

teancom:

Server 2.10 splits music/recommendations into a rows listing without embedded items plus a per-row music/recommendations/items command. The home page filtered out rows without playable items, so it rendered empty against 2.10 (beta) servers. Older servers keep embedding items.

MediaItemRepository.recommendationRows() bridges both response shapes as one snapshot stream: embedded rows pass through as a single resolved snapshot; item-less rows emit an immediate all-loading snapshot, then resolve per row in parallel. The first row is probed alone so an old server that returned only empty rows costs at most one failing RPC (RPC errors surface to the user as toasts). CarPlay consumes the same flow through a one-shot merged variant.

The home screen renders each row through CategoryRow's DataState overload: shimmer placeholders while a row's items load, auto-hide for rows that resolve empty, so the slowest provider no longer gates the whole page. reconcileHomeRows is generic over the row type, letting loading rows reconcile by identity before their items arrive. Shortcuts load concurrently instead of gating on recommendations.

Covered by RecommendationFoldersCompatTest (both server shapes, probe behavior, streaming snapshots, failure paths) and HomeScreenCategoriesTest (row visibility rules).

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
@@ -11,9 +11,11 @@ import io.music_assistant.client.data.model.server.ServerMediaItem
 import io.music_assistant.client.data.model.server.events.MediaItemAddedEvent
 import io.music_assistant.client.data.model.server.events.MediaItemDeletedEvent
 import io.music_assistant.client.data.model.server.events.MediaItemUpdatedEvent
+import io.music_assistant.client.utils.HasConnectionData
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.IO
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
@@ -25,6 +27,7 @@ import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.scan
 import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.withContext
 
 /**
  * Single seam between RPC + DTO land and the UI's typed `AppMediaItem` world.
@@ -69,42 +72,50 @@ class MediaItemRepository(
 
     /**
      * The home-page recommendation rows as the server returned them, with or
-     * without embedded items (see [needPerRowItemFetch]).
+     * without embedded items (see [supportsRecommendationRowItems]).
      */
     suspend fun fetchRecommendationRows(): Result<List<RecommendationFolder>> =
-        fetchMediaItems(Request.Library.recommendations())
-            .map { items -> items.filterIsInstance<RecommendationFolder>() }
+        withContext(Dispatchers.IO) {
+            fetchMediaItems(Request.Library.recommendations())
+                .map { items -> items.filterIsInstance<RecommendationFolder>() }
+        }
+
+    /**
+     * Whether the connected server strips the items from `music/recommendations`
+     * rows and serves each row's contents via `music/recommendations/items`
+     * instead. This can be simplified once v2.10 is our minimum supported server
+     * version.
+     */
+    fun supportsRecommendationRowItems(): Boolean =
+        (apiClient.sessionState.value as? HasConnectionData)?.serverInfo?.schemaVersion
+            ?.let { it >= RECOMMENDATION_ITEMS_SCHEMA } == true
 
     /**
      * One-shot, fully-resolved recommendation rows for consumers without a
-     * progressive UI (e.g. CarPlay lists). The first row is probed alone —
-     * see [needPerRowItemFetch].
+     * progressive UI (e.g. CarPlay lists).
      */
     suspend fun fetchRecommendationFolders(): Result<List<RecommendationFolder>> {
         val folders = fetchRecommendationRows().getOrElse { error ->
             if (error is CancellationException) throw error
             return Result.failure(error)
         }
-        if (!folders.needPerRowItemFetch()) return Result.success(folders)
+        if (!supportsRecommendationRowItems()) return Result.success(folders)
 
-        val firstRowItems = fetchRecommendationRowItems(folders.first())
-            ?: return Result.success(folders)
-        val remaining = coroutineScope {
-            folders.drop(1).map { folder ->
-                async {
-                    folder.copy(items = fetchRecommendationRowItems(folder).orEmpty())
-                }
-            }.awaitAll()
-        }
         return Result.success(
-            listOf(folders.first().copy(items = firstRowItems)) + remaining,
+            coroutineScope {
+                folders.map { folder ->
+                    async {
+                        folder.copy(items = fetchRecommendationRowItems(folder).orEmpty())
+                    }
+                }.awaitAll()
+            },
         )
     }
 
     /** Items of one recommendation row, or null when the fetch failed (logged). */
     suspend fun fetchRecommendationRowItems(
         folder: RecommendationFolder,
-    ): List<AppMediaItem>? =
+    ): List<AppMediaItem>? = withContext(Dispatchers.IO) {
         fetchMediaItems(
             Request.Library.recommendationItems(folder.provider, folder.itemId),
         ).getOrElse { error ->
@@ -115,6 +126,7 @@ class MediaItemRepository(
             )
             null
         }
+    }
 
     /**
      * Issue [request] and decode its payload as a single client media item.
@@ -189,14 +201,5 @@ class MediaItemRepository(
 private fun <T : AppMediaItem> List<T>.replacing(changed: T): List<T> =
     map { if (it.itemId == changed.itemId) changed else it }
 
-/**
- * True when a `music/recommendations` response is the item-less shape (2.10+):
- * every row arrived without items, so each row's contents must be fetched via
- * [MediaItemRepository.fetchRecommendationRowItems]. A server that embeds items
- * always populates at least the rows worth showing. Callers should probe the
- * first row alone before fanning out: a pre-2.10 server that legitimately
- * returned only empty rows lacks the items command, and every failing call
- * there surfaces a user-visible error toast.
- */
-fun List<RecommendationFolder>.needPerRowItemFetch(): Boolean =
-    isNotEmpty() && all { it.items.isNullOrEmpty() }
+/** Server schema version that split `music/recommendations` into rows + per-row items. */
+private const val RECOMMENDATION_ITEMS_SCHEMA = 39
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
@@ -15,20 +15,16 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.SharingStarted
-import kotlinx.coroutines.flow.channelFlow
-import kotlinx.coroutines.flow.last
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.scan
 import kotlinx.coroutines.flow.shareIn
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 
 /**
  * Single seam between RPC + DTO land and the UI's typed `AppMediaItem` world.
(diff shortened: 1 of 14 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -61,6 +71,82 @@ class MediaItemRepository(
...
+            .getOrThrow()
+            .filterIsInstance<RecommendationFolder>()
+        val itemLess = folders.isNotEmpty() && folders.all { it.items.isNullOrEmpty() }
+        if (!itemLess) {
+            send(folders.map { RecommendationRow(it, itemsLoading = false) })
+            return@channelFlow
+        }
+
+        val rows = MutableList(folders.size) { RecommendationRow(folders[it], itemsLoading = true) }
+        val lock = Mutex()
```
formatBCE: I don't feel like we need lock here, as well as channelFlow. It's unnecessary over-complication.
 This piece of code should be just coroutine block, that fires rows with loading state to stateFlow, and then launches loading jobs for all rows (that are visible).

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeRowsConfig.kt
```diff
@@ -17,24 +16,25 @@ import io.music_assistant.client.ui.compose.common.items.ItemCategory
  *   from config keep server order and sort after the known ones (stable sort).
  * - Stored ids no longer present on the server are ignored.
  */
-internal fun reconcileHomeRows(
-    categories: List<ItemCategory<*>>,
+internal fun <T> reconcileHomeRows(
+    rows: List<T>,
```
formatBCE, line 25: Avoid using generic <T> types if there's similar functionality. In this case you have `id` for both - use it as interface field (e.g. `IdProvider`), and declare that as type.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -207,38 +209,36 @@ class HomeScreenViewModel(
         }
 
         loadDataJob = viewModelScope.launch {
-            _state.update {
-                it.copy(
```
formatBCE, line 211: This (initial loading state) is missing?

formatBCE (commented):
I feel like it's more iOS code than Kotlin hahaha :)

@seadowg what do you think?

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
teancom: So this is actually because of CarPlay (speaking of!), which doesn't have progressive loading of templates so it needs to be one-shot, even on the new server. So we use the channelFlow via .last() which means it needs to complete, which StateFlow doesn't.

And the lock isn't for the channel, it's the parallel fetches that all mutate the shared rows list. So each emitted snapshot is consistent.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
teancom: This was a duplicate, we used to set it twice. It's still there a few lines up.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeRowsConfig.kt
teancom: Done, though I made it bounded-generic so the callers didn't have to cast the result pairs.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
seadowg: Yeah, I agree with @formatBCE that this feels overly complex. I think `HomeScreenViewModel` could fetch the recommendation folder list and then launch individual jobs as a list for each one (exposing this as a `List<DataState>` to the UI).

If we need to load recommendation as a "one-shot" for CarPlay, we can have specific code for that which would be easy enough to create out of the above parts (just fire off a bunch of requests and `join` on the jobs or something).

composeApp/src/commonTest/kotlin/io/music_assistant/client/data/repository/RecommendationFoldersCompatTest.kt
```diff
@@ -0,0 +1,262 @@
...
+import kotlin.test.assertTrue
+import kotlin.test.fail
+
+/**
+ * [MediaItemRepository.fetchRecommendationFolders] must bridge both shapes of
+ * the `music/recommendations` response: rows with their items embedded (servers
+ * before 2.10), and item-less rows (2.10+) whose contents come from a per-row
+ * `music/recommendations/items` call.
+ */
+class RecommendationFoldersCompatTest {
```
seadowg, line 34: Instead of here, this could be tested in `HomeScreenTest` - we already have a mechanism for simulating different API behaviours in `FakeServiceClient` (`setLegacyVersion`) that is used in `ShortcutsTest` and `AudioChainTest`.

seadowg (changes requested):

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
teancom: Fixed, thanks for the pointers (both of you) - I had my blinders on and wasn't looking "up".

teancom: We'll have to wait out this GitHub weather to prove it in CI, but I swear it works on my machine. 😏

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -210,15 +211,65 @@ class HomeScreenViewModel(
...
+        val probeItems = mediaItemRepository.fetchRecommendationRowItems(folders.first())
+        if (probeItems == null) {
+            setRecommendationRows(
+                folders.map { RecommendationRowState(it, DataState.Data(emptyList())) },
+            )
+            return
+        }
+        setRowItems(folders.first(), probeItems)
+        coroutineScope {
+            folders.drop(1).forEach { folder ->
```
formatBCE: What does this drop do? I don't like when logic is built on indexes... It's prone to errors.
(I read the comment, but I don't really like this logic at all).

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -221,11 +190,13 @@ private fun <T : AppMediaItem> List<T>.replacing(changed: T): List<T> =
...
+ * always populates at least the rows worth showing. Callers should probe the
+ * first row alone before fanning out: a pre-2.10 server that legitimately
+ * returned only empty rows lacks the items command, and every failing call
+ * there surfaces a user-visible error toast.
  */
-data class RecommendationRow(
-    val folder: RecommendationFolder,
-    val itemsLoading: Boolean,
-)
+fun List<RecommendationFolder>.needPerRowItemFetch(): Boolean =
```
formatBCE: I would actually gate this under schema version. Otherwise we will be trying it for every empty row even if server doesn't have it.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
```diff
@@ -210,15 +211,65 @@ class HomeScreenViewModel(
...
+        setRecommendationRows(folders.map { RecommendationRowState(it, DataState.Loading()) })
+        val probeItems = mediaItemRepository.fetchRecommendationRowItems(folders.first())
+        if (probeItems == null) {
+            setRecommendationRows(
+                folders.map { RecommendationRowState(it, DataState.Data(emptyList())) },
+            )
+            return
+        }
+        setRowItems(folders.first(), probeItems)
+        coroutineScope {
```
formatBCE, line 235: This means every call to server will be done in main (UI) thread. 
`coroutineScope` inside of `viewModelScope` is using same Dispatcher, as VM - and this is UI thread (Dispatchers.Main).

So this coroutine scope doesn't do anything.
What you should use is
`withContext(Dispatchers.IO) {`

This will send network requests to the IO dispatcher, dedicated for this.
Same with other network requests, actually...
We either do it here, or wrap them into IO dispatcher in repository. 

It's usually overlooked by me and others, while there are not so many requests - but when it's like this, 20-30 rows each loading the items - it bites. We need to start doing this.

formatBCE (changes requested):

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/HomeScreenViewModel.kt
formatBCE: BTW not only this doesn't help with network requests, but it also unbinds from `viewModelScope`, which means that requests will be executed and not canceled even if ViewModel is dismantled (e.g. user exits the screen, but network is still trying to fetch those and deliver).

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
teancom: To clarify, you mean to explicitly check something like 'if LOCAL_SCHEMA_VERSION < 39' (with, you know, consts and stuff). And maybe a note to pull this out once our minimum supported server version is 2.10?

formatBCE: No, server schema. :)

teancom: Sorry, right, grabbed the wrong variable. But if I s/LOCAL_SCHEMA_VERSION/server version/, that's what you're looking for?

formatBCE: Yeah, I mean, if schema doesn't support rows loading, we should just omit it.

teancom:
Alright, I believe I addressed all of the (excellent) feedback. Please take another look, very happy to iterate more if need be. I'll also be living-on testing this on my phone.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -69,42 +72,50 @@ class MediaItemRepository(
...
-                async {
-                    folder.copy(items = fetchRecommendationRowItems(folder).orEmpty())
-                }
-            }.awaitAll()
-        }
+        if (!supportsRecommendationRowItems()) return Result.success(folders)
+
         return Result.success(
-            listOf(folders.first().copy(items = firstRowItems)) + remaining,
+            coroutineScope {
```
formatBCE, line 105: I feel like this should be done differently (cancelable scope). But okay, let's merge, and I will work on it a bit :)

formatBCE (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -61,6 +70,64 @@ class MediaItemRepository(
         }
     }
 
+    /**
+     * The home-page recommendation rows as the server returned them, with or
+     * without embedded items (see [supportsRecommendationRowItems]).
+     */
+    suspend fun fetchRecommendationRows(): Result<List<RecommendationFolder>> =
+        withContext(Dispatchers.IO) {
```
seadowg, line 78: Does this need to use IO? I've still not actually dug into if the commands call/response is handled off the UI thread or not (although I assume it is).

composeApp/src/commonTest/kotlin/io/music_assistant/client/data/repository/RecommendationFoldersCompatTest.kt
seadowg: Is this test still needed?

seadowg (changes requested):

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
formatBCE: It for sure needs to use IO.
If it's received on UI thread, it will be collected from flow by coroutine running in viewModelScope with Main dispatcher. But the network request itself should be taken out of Main - because it introduces lags on UI. IO is for network, Main is for Compose updates.

seadowg: > But the network request itself should be taken out of Main - because it introduces lags on UI. IO is for network, Main is for Compose updates.

Agreed with that, but does that not mean other places we call `sendRequest` should also be using `Dispatchers.IO`? I'd thought we ended up dispatching with `IO` for the actual network interaction further down as all the `sendRequest` calls happen on `Main`. Something like this as an example: 

https://github.com/music-assistant/mobile-app/blob/2df051e8fb30d54fa760e79ae55bd118b2f36012/composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemDetailsViewModel.kt#L151

I might just need to dig down and read the `ServiceClient` implementation code some more to understand.

composeApp/src/commonTest/kotlin/io/music_assistant/client/data/repository/RecommendationFoldersCompatTest.kt
teancom: Half of it is, trimmed the duplicated half. 👍

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
formatBCE: Yes, that's exactly what I wrote in my previous comments. Once we're getting more and more from network, we need to sweep our dispatchers usage, and actually use scoped repositories that run on their own injected dispatchers. That's how all good apps are done.

teancom: I'm starting to get confused (starting? hah!) - is there something else I should be doing here? I could adopt injected dispatchers repo-wide (and hell, enable the detekt rule to make it mandatory) in this PR, if we want. Or I can do that in a separate PR. Or leave that to you two. 😅

formatBCE: @teancom no, nothing to do for you - we need to sweep all the code we have, but it's scope of other task.

seadowg: > Yes, that's exactly what I wrote in my previous comments. Once we're getting more and more from network, we need to sweep our dispatchers usage

Ah gotcha. I thought you'd meant this one case, but you mean everything - it's the inconsistency I was confused about.

> That's how all good apps are done.

Agreed! I'm surprised we haven't experienced more ANRs or crashes, but it looks like the actual network access still happens on `IO` (in `KtorServiceClient#sendRequestRaw`), but we're likely to run into issues (as you say) with pretty much all the work before/after the actual access still sitting on `Main` - and it's just simpler in my mind to have the request call stack all running in `IO`. 

But yes, not something to fix here @teancom! Just something I realized I hadn't really thought much about for this app's structure.

seadowg (approved):

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
formatBCE: Yup that's because pure web sockets are fast, and we ain't pullin' nothing serious so far.

PR #846, opened 2026-08-06
Add sort to "View all" screens

### LINKED ISSUE #592: Scrolling to top when already on top in item details breaks theming (overscroll)
## Context
 - Operating system and version: iOS 26.5.1
 - App version: iOS TestFlight 1.0 (35)
 - Music Assistant server version: 2.9.0 Stable

### Problem description

Scrolling to the top when already on top in item details breaks theming. iOS performs the 'bouncy/sticky' scroll, allowing the elements (topbar and item details section) to be separated, showing an empty background.

### Expected behavior

Disallow scrolling up when already on top to make sure elements don't separate (make it sticky?).

Image

seadowg:
Work towards https://github.com/music-assistant/mobile-app/issues/801
Follow up for https://github.com/music-assistant/mobile-app/pull/834 (more on the way)
Closes #592

Like with view mode, I've extracted a new component specifically for sorting items. I've not done a lot to make sorting easier to add as we'll ideally not be needing as many client-side sorted lists in the future.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/NoOverscroll.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.compose.foundation.LocalOverscrollFactory
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+
+// Drop the list's inner overscroll so a downward pull at the top reaches the
+// ancestor collapse NestedScrollConnection instead of being eaten by the iOS
+// Cupertino rubber-band. Covers both the populated LazyColumn and the empty
+// verticalScroll Column.
+@Composable
+fun NoOverscroll(content: @Composable () -> Unit) {
```
formatBCE, line 12: Yeeeee!

composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
```diff
@@ -67,3 +72,14 @@ fun Modifier.conditional(
 }
 
 inline fun <reified T : Any> Result<Answer>.resultAs(): T? = getOrNull()?.resultAs()
+
+fun <T1, T2, R> CoroutineScope.combineAsStateFlow(
```
formatBCE, line 76: While I love utility methods, I think it doesn't bring that much convenience comparing to regular `combine(f1, f2, transform)`, but adds unnecessary wrapping.

formatBCE (approved):
Approving, but have one comment there.  Everything else is great.

composeApp/src/commonMain/kotlin/io/music_assistant/client/utils/Ext.kt
seadowg: Not wrapping means you need to define the "initial value" (because `combine` is for `Flow` rather than `StateFlow`) though which always reads strangely and is pretty fragile. This extension does that work for you so that the initial value is always guaranteed to be `transform(flow1.value, flow2.value)`.

formatBCE: Ok, fair enough - I just don't see many usages of this.

seadowg: Yeah, that's potentially true. I could just make it private in `ItemListViewModel.kt`?

formatBCE: Leave it. Probably we'll have another usage. :)

PR #868, opened 2026-08-11
Add CarPlay play from here support

### LINKED ISSUE #867: CarPlay: support Play from here in album and playlist drilldowns
## Problem

CarPlay does not currently offer **Play from here** when a track is selected inside an album or playlist drilldown. The car settings already include `PLAY_FROM_HERE`, and Android Auto supports this behavior, but CarPlay filters it out because it does not preserve the parent container context.

## Solution

- Offer **Play from here** for CarPlay track default-tap settings.
- In album and playlist drilldowns, dispatch the parent media URI with the tapped track item ID as `start_item`.
- If the parent context is unavailable, fall back to playing the tapped track by itself.

Reconnect and session restoration are intentionally outside the scope of this change.

dandckr-ops:
Closes #867

## Summary
- enable `PLAY_FROM_HERE` for CarPlay track taps
- preserve the album or playlist parent through CarPlay drilldown handlers
- dispatch the parent media URI with the tapped track ID as `start_item`
- fall back to replacing the queue with the tapped track when parent/start-item context is missing, matching Android Auto

Reconnect/session restoration is intentionally excluded from this PR.

## Test-driven development
- changed the missing-context planner expectations first and observed the expected failures
- added planner coverage for album and playlist parents, missing parent, missing start item, and unchanged plain-action dispatch

## Validation
- `CI=true ./gradlew detektAll`
- `./gradlew :composeApp:testAndroidHostTest`
- `./gradlew :androidApp:testDebug`
- `./gradlew :androidApp:assembleDebug`
- `./gradlew :androidApp:lintDebug`
- independent read-only diff review: passed with no security or logic errors

The existing PR workflow will run the Kotlin/Native test target and unsigned iOS simulator compile on macOS.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
@@ -44,9 +44,9 @@ fun DefaultClickOption.isCarSupported(platform: CarPlatform, kind: ItemKind): Bo
 
     return if (this == DefaultClickOption.PLAY_FROM_HERE) {
         // Needs a parent container + start track, only resolvable for a TRACK tap inside an
-        // album/playlist drilldown — and only AA threads that context through (see AutoLibrary).
+        // album/playlist drilldown. Android Auto and CarPlay both thread that parent context.
         // Tap-only: TRACK is never a browsable bulk kind, so this also keeps it out of bulk lists.
-        platform == CarPlatform.ANDROID_AUTO && kind == ItemKind.TRACK
+        kind == ItemKind.TRACK
     } else {
         when (platform) {
             CarPlatform.ANDROID_AUTO -> true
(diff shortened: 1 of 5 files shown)


formatBCE:
@teancom can you check?

composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
```diff
@@ -69,6 +69,45 @@ fun Map<ItemKind, List<DefaultClickOption>>.carBulkActions(
...
+    val mediaUris: List<String>,
+    val option: QueueOption,
+    val radioMode: Boolean,
+    val startItem: String? = null,
+)
+
+/**
+ * Resolve a configured car tap into an MA play-media payload. PLAY_FROM_HERE targets the parent
+ * album/playlist URI and starts at the tapped track; plain actions target the tapped item's URI.
+ * Returns null instead of degrading PLAY_FROM_HERE to single-track playback when context is absent.
```
teancom: While there is no way to get a context-less tapped playback, at least as we are today, I'd rather we failed open (played the track the user tapped) vs. silently fail to play the track, in the event that does happen (bug, or new path in the future). That also matches how we've configured Android Auto which *does* have context-less playback normally.

teancom (changes requested):
Everything else looks great! I tested on my phone+CarPlay, worked like a charm.

composeApp/src/commonMain/kotlin/io/music_assistant/client/settings/CarActionPrefs.kt
dandckr-ops: Updated in `5e50fc9`: when the parent or start-item context is missing, `PLAY_FROM_HERE` now falls back to replacing the queue with the tapped track, matching Android Auto. I also added tests for both missing-parent and missing-start-item cases.

teancom (approved):
Thank you! Looks good to me.

formatBCE:
Thank you @dandckr-ops !

PR #872, opened 2026-08-12
Fix CarPlay authoritative state refresh

dandckr-ops:
## Summary
- refresh authoritative player and queue state when CarPlay attaches
- preserve newer event/optimistic queue state when a delayed full queue snapshot arrives
- use the stale-safe merged snapshot for position tracking and local-player forwarding

## Scope
- read-only refresh only; no playback, seek, or queue mutation added
- no Sendspin reconnect, route-loss, or playback-authorization changes
- no Android behavior changes

## Validation
- complete Android-host test target passed (296 tests, 0 failures)
- focused CarPlay queue snapshot merge and existing Now Playing/remote-command tests passed
- `:composeApp:compileKotlinIosSimulatorArm64` passed
- `git diff --check` passed
- blocker-only conformity review passed

## Physical validation
Physical iPhone/CarPlay validation was not performed because no Apple Developer signing path is available. This remains explicitly unverified.

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
@@ -76,6 +76,18 @@ import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlin.coroutines.CoroutineContext
 
+/** Preserve newer event/optimistic state when a delayed full snapshot arrives. */
+internal fun mergeFullQueueSnapshot(
+    retained: List<QueueInfo>,
+    incoming: List<QueueInfo>,
+): List<QueueInfo> {
+    val retainedById = retained.associateBy { it.id }
+    return incoming.map { candidate ->
+        val current = retainedById[candidate.id]
+        if (current != null && candidate.isBefore(current)) current else candidate
+    }
+}
+
 @OptIn(FlowPreview::class)
 class MainDataSource(
     private val settings: SettingsRepository,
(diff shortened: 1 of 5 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
```diff
@@ -923,6 +935,15 @@ class MainDataSource(
         _userSelectedPlayerId.update { player.id }
     }
 
+    /**
+     * Re-read authoritative player/queue state when CarPlay attaches.
+     * This reuses the existing read-only refresh path and never issues playback commands.
+     */
+    fun refreshCarPlayNowPlayingState() {
```
formatBCE: I don't like platform-specific code slipping into common code. If you search, you won't find AndroidAuto or CarPlay code in common. Let's keep it like this.

formatBCE (changes requested):

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/MainDataSource.kt
dandckr-ops: Good point. Fixed in `34d1565`: the common-layer API, documentation, logging, and snapshot-merge test now use platform-neutral naming. The CarPlay-specific name remains only in the iOS bridge.

formatBCE (approved):

formatBCE (approved):

formatBCE:
LGTM, thank you!

PR #876, opened 2026-08-14
Fix flaky test synchronization

teancom:

Who wrote this crappy code??? *quick git blame session later* You know what, there's no need to point fingers, I'm just happy to be able to help contribute a fix.

### DIFF
--- a/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
+++ b/androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
@@ -49,7 +49,7 @@ import kotlinx.serialization.json.encodeToJsonElement
 
 class FakeServiceClient : ServiceClient {
     private var legacyVersion: LegacyVersion? = null
-    private var requestErrors: Boolean = false
+    private val requestErrors = java.util.concurrent.atomic.AtomicBoolean(false)
     private var connectionError: Exception? = null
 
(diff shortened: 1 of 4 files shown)


formatBCE (approved):
I don't generally like mutex3s - they're for low-level operations usually. But okay. :)

teancom:
> I don't generally like mutex3s - they're for low-level operations usually. But okay. :)

Yeah, I was trying to not just take the "keep adding more retries and longer timeouts" approach but admittedly I'm not being as persnickety about the code itself since it's "just" tests. As long as they don't flake and they still test what they're supposed to test... 🤷

androidApp/src/test/kotlin/io/music_assistant/client/support/FakeServiceClient.kt
```diff
@@ -49,7 +49,7 @@ import kotlinx.serialization.json.encodeToJsonElement
 
 class FakeServiceClient : ServiceClient {
     private var legacyVersion: LegacyVersion? = null
-    private var requestErrors: Boolean = false
+    private val requestErrors = java.util.concurrent.atomic.AtomicBoolean(false)
```
formatBCE: Import?

teancom: Doh.

formatBCE (approved):

PR #884, opened 2026-08-19
Add the Sendspin Noise KKpsk2 core, device identity, and PSK trust store

teancom:
I have the new Sendspin encryption working. For review-ability I've split it into three PRs, of which this is the first. It's "just" the encryption bits, not actually wired into anything. The second commit refactors the connection establishing bits into protocol sessions while still only supporting the legacy approach, and the third finishes by wiring it all together.

- noise: Noise Protocol Framework state machines for Noise_KKpsk2_25519_ChaChaPoly_SHA256, pinned by cacophony reference vectors; the Sendspin handshake driver (init exchange, raw-byte prologue, psk_id candidate selection, stored-pubkey post-match); message-type framing and fragmentation; psk_id derivation and sentinel constants; crypto primitives behind a swappable NoiseCrypto interface, backed by cryptography-kotlin (JCA on Android, CryptoKit on iOS).
- identity: persisted X25519 device identity + versioned single-blob trust store (Pairing PSK, long-term records, config flags, live PSK-candidate set) over the app's regular settings storage, with clean regeneration on missing or corrupt data.
- pairing/PairingToken: the out-of-band token codec (base32, 2->9 transliteration, canonical-bits validation) with the spec vector.
- model/Messages: encrypted-protocol message models alongside the frozen legacy models; LegacyWireGoldenTest pins the legacy wire bytes.
- 

I'm marking this as a Chore so it doesn't go into the changelog. I figured we add the third and final PR into the changelog with a simple 'Enable Sendspin encryption' title, since that's what users will care about.

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -85,6 +85,9 @@ kotlin {
             implementation(libs.androidx.media)
             implementation(libs.androidx.browser)
 
+            // Noise primitives backend (JCA crypto via the JDK provider).
+            implementation(libs.cryptography.provider.jdk)
+
             implementation(libs.coil)
             implementation(libs.concentus)
         }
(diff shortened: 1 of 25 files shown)


formatBCE (approved):
Nice one @teancom !

PR #885, opened 2026-08-19
Refactor Sendspin connection establishment into protocol sessions

teancom:

Restructures the transport/connection layer so a second (encrypted) protocol can plug in beside the legacy one; the legacy wire remains byte-identical (pinned by tests) and no encrypted code path is reachable yet.

- SendspinTransport now emits one lossless, source-ordered, epoch-carrying event stream; both transports, SendspinWsHandler, and DataChannelWrapper (single ordered inbound flow + injectable receive source) adapted.
- SendspinProtocolSession owns establishment as the transport's sole collector: LegacySession extracts today's auth/hello behavior; audio is demuxed by the session; the client reacts to session lifecycle events only (auto-resume re-homed to post-reconnect readiness).
- MessageDispatcher consumes the session's ordered pump and sends through the session's serialized sender; awaitInitialOutcome() gives the factory a definitive attach outcome, and WebRTCChannelGate pins the single-use channel semantics (failed attaches exhaust the wrapper too).

PR two of three and frankly the one that needs the most scrutiny. "It works on my phone" is great and all, but this is a pretty big refactoring. I would be *shocked* if I hadn't missed something.


composeApp/src/commonMain/kotlin/io/music_assistant/client/webrtc/WebRTCConnectionManager.kt
```diff
@@ -481,8 +481,12 @@ class WebRTCConnectionManager(
         // Collect incoming messages from the flow
         messageListenerJob = scope.launch {
             try {
-                channel.messages.collect { msg ->
-                    _incomingMessages.emit(msg)
+                // Sole collector of the ma-api channel's ordered inbound
+                // stream; this channel carries text RPC frames only.
+                channel.inbound.collect { msg ->
```
formatBCE, line 486: `channel.inbound.filterNotNull { it as?  DataChannelInbound.Text}.collect { _incomingMessages.emit(it) }`

formatBCE (approved):
Woof, huge. LGTM.

composeApp/src/commonMain/kotlin/io/music_assistant/client/webrtc/WebRTCConnectionManager.kt
teancom: I'll take a look at doing this in the third PR, but I'm afk and won't be doing that until this evening.

PR #912, opened 2026-08-24
Add support for new shared icons

### LINKED ISSUE #756: Support MA player icons
As discussed [on Discord](https://discord.com/channels/753947050995089438/1523995860877312070), MA will support a new standardized set of icons for players in the future that will be available as a set of downloadable SVGs named after the IDs that the server will return for them (i.e. `turntable` will map to `turntable.svg`).

The app should support these by including vector assets converted from the SVGs and mapping them accordingly.

seadowg:
Closes https://github.com/music-assistant/mobile-app/issues/756

This includes the icons as a new shared KMP module which is unusual for the project, but pretty normal in Gradle based projects in my experience for "mini frameworks" like this. I find this approach especially useful for specific sets of resources as otherwise they end up in one big pile.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayerIcon.kt
```diff
@@ -27,17 +24,19 @@ fun PlayerIcon(
...
-        return
-    }
-    val fallback = if (player.isGroup) SpeakerMultipleIcon else Icons.Default.Speaker
     MdiIcon(
         name = player.icon,
         modifier = modifier,
         tint = tint,
         fallback = {
-            Icon(fallback, contentDescription = null, modifier = modifier, tint = tint)
+            val iconId = if (isLocal) SharedIcons.SMARTPHONE else player.icon
```
seadowg, line 32: I considered just dropping support for `mdi-*` icons entirely, but that felt overly aggressive. We can do that when we move towards supporting server 2.11 which will result in us dropping official support for 2.9.

shared-icons/src/commonTest/kotlin/io/music_assistant/client/sharedicons/SharedIconsTest.kt
```diff
@@ -0,0 +1,73 @@
...
+
+class SharedIconsTest {
+    @Test
+    fun `getResource returns speaker for unknown ID`() {
+        val resource = SharedIcons.getResource("unknown")
+        assertEquals(Res.drawable.speaker, resource)
+    }
+
+    @Test
+    fun `getResource returns resource for every ID in manifest`() {
```
seadowg, line 20: The coverage feels pretty low here, but was motivated by the main fear I had when working on this: "I hope I haven't missed one of these". I think this should give someone updating the icons in future a little more confidence while they're doing it.

shared-icons/build.gradle.kts
```diff
@@ -0,0 +1,86 @@
+plugins {
+    alias(libs.plugins.kotlinMultiplatform)
+    alias(libs.plugins.androidMultiplatformLibrary)
+    alias(libs.plugins.composeMultiplatform)
+    alias(libs.plugins.composeCompiler)
+}
+
+kotlin {
```
seadowg, line 8: Other than adding things the module needed, the only change I've made here to the boilerplate generated by Android Studio for a new KMP module is that I removed the Android device test config (which we won't need here). Sadly, it's still pretty noisy! That's not unlike the default Android module boilerplate honestly, but I less confidence chopping bits of this out.

formatBCE (approved):
Nice!

PR #923, opened 2026-08-26
Fix player-switching crash by keeping MdiIcon structure stable

teancom:

Crash report (0.12.0, Android, during player selection/switching):

    java.lang.IllegalArgumentException: Cannot disable reuse from root
    if it was caused by other groups

Diagnosis:

- Thrown by the node-reuse invariant in androidx.compose.runtime 1.11.2 (GapComposer/LinkComposer.endReuseFromRoot) during a main-thread recomposition.
- Google fixed only the pausable-composition variant of this crash (b/488433633, runtime 1.10.6 / 1.11.0-alpha01+), which 1.11.2 already contains. Release notes for 1.11.3, 1.11.4 and 1.12.0 list no further fixes for this crash family, so upgrading is not a fix.
- The shipped trigger profile matches the player icon stack: MdiIcon early-returned and swapped its emitted tree between two structurally different shapes (a plain fallback Icon vs BoxWithConstraints { Text }). That flip happens asynchronously (the one-time MDI codepoint table load) and per player (null vs non-null icon names), inside keyed lazy content whose compositions Compose reuses and reorders exactly while the user selects or switches players: the HorizontalPager keyed by player id and the keyed, reorderable select-player dialog. The crash log shows the players list wholesale-replaced twice during a reconnect while this was happening.

Fix:

- MdiIcon with a non-null fallback now always emits the same group structure: a BoxWithConstraints root whose content switches between the glyph and the fallback. No early return, so a reused composition never sees its root node kind change.
- The null-fallback mode keeps the legacy glyph-or-nothing shape so provider icons never occupy layout space for unknown names; it is not used in keyed lazy content.
- PlayerIcon fallback fills the container instead of re-applying the caller modifier.

Test:

- PlayerIconPagerReuseTest is a soak test, NOT a bug regression test: verified it also passes against the pre-fix code, both as a plain harness and as a frame-interleaved variant (manual clock, wholesale player replacement, programmatic page jumps) - Robolectric cannot reach the race. It keeps automated coverage on the icon render path under keyed-lazy reuse churn and would catch any deterministically triggerable crash there; a true repro likely needs an instrumented device run.

Confidence:

- No synthetic repro tripped the invariant in isolation (timing dependent), so this is strongly evidenced hardening of the exact risky pattern rather than a proven causal fix. If the crash persists, the next suspect is the conditional badge slots in PlayerBadgesRow (also new in 0.12.0).


androidApp/src/test/kotlin/io/music_assistant/client/ui/compose/home/players/PlayerIconPagerReuseTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.koin.dsl.module
+
+/**
+ * Regression guard for the 0.12.0 player-switching crash (Compose node-reuse invariant
+ * failure). Churns keyed pager pages — reorder, remove, add, swipes — around the real
+ * [PlayerIcon] stack while MDI glyph resolution flips asynchronously, mirroring player
+ * selection/switching. Full diagnosis lives in the fix commit message.
+ */
+@RunWith(AndroidJUnit4::class)
+class PlayerIconPagerReuseTest {
```
seadowg: This test doesn't fail with the previous `MdiIcon` implementation - I don't think there's much value in including it.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayerIcon.kt
```diff
@@ -31,10 +31,11 @@ fun PlayerIcon(
         fallback = {
             val iconId = if (isLocal) SharedIcons.SMARTPHONE else player.icon
             val iconRes = SharedIcons.getResource(iconId)
+            // Fills MdiIcon's container, which already carries `modifier`; re-applying
```
seadowg: This comment describes the change: it won't make much sense to someone reading the code instead of a diff.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/providers/MdiIcon.kt
```diff
@@ -60,21 +60,29 @@ fun codePointToString(codePoint: Int): String =
...
     modifier: Modifier = Modifier,
     tint: Color = Color.White,
-    fallback: @Composable () -> Unit = {},
+    fallback: (@Composable () -> Unit)? = null,
 ) {
     val table by koinInject<MdiCodepoints>().table.collectAsStateWithLifecycle()
     val codePoint = name?.let { table[normalizeMdiName(it)] }
-    if (codePoint == null) {
-        fallback()
+    if (codePoint == null && fallback == null) {
```
seadowg: How about we just guard the `BoxWithConstraints` instead of using an early return here? I think that'd be easier to follow.

seadowg (changes requested):
It seems very strange that the previous code would cause an issue, but I can definitely believe there's some Compose weirdness at play here given the exception. I like the new structure of `MdiIcon` so I don't think there's much harm in making the change.

formatBCE (commented):
My robot said basically the same. But I asked it to make bug reproducible on the old code, and he couldn't. So I discarded its first guess. Then it started to think more and got out of tokens. Waiting for session to restart. :)

androidApp/src/test/kotlin/io/music_assistant/client/ui/compose/home/players/PlayerIconPagerReuseTest.kt
teancom: I was really on the fence about including it, happy to pull it.

PR #935, opened 2026-08-27
Label player controls

seadowg:
As discussed on Discord.

Screenshot_20260827_142602

I've also:

- Removed the duplicated items from the options menu. We might want to move some of these back in to the options menu or offer some kind of customization around this later, but for now I think it's best to just have them in one place to keep the options menu as clean as possible.
- Remove unused strings. We don't get a check for this in CMP like we get with Android projects sadly and Detekt is only intended for linting Kotlin files which means extending that probably isn't answer. For now, it's a good candidate as a task to delegate to your favourite agent - I reckon we will see better tooling around this in the future.


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/PlayerBadgesRow.kt
```diff
@@ -207,7 +215,7 @@ internal fun BadgePill(
             .background(container)
             .then(onClick?.let { Modifier.clickable(onClick = it) } ?: Modifier)
             .padding(horizontal = 8.dp, vertical = 4.dp)
-            .semantics(mergeDescendants = true) { this.contentDescription = contentDescription },
+            .clearAndSetSemantics { this.contentDescription = contentDescription },
```
seadowg, line 218: This will stop badge text interfering with the content description for these buttons and add noise to screen readers.

composeApp/src/commonMain/composeResources/values-de/strings.xml
formatBCE, line 1: I feel like Lokalise would clean that up - but that's good too. :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/home/players/SleepTimerBadge.kt
```diff
@@ -80,7 +81,11 @@ fun SleepTimerBadge(
             modifier = Modifier.size(BADGE_ICON_SIZE),
         )
         Text(
-            text = remaining.formatDuration(),
+            text = if (remaining != null) {
```
formatBCE, line 84: `text = remaining?.formatDuration() ?: stringResource(Res.string.player_sleep_timer)`

formatBCE (approved):
Approved with nit pick :)

PR #946, opened 2026-08-30
Extract artist specific code from item details screen

seadowg:
Work towards #801

The initial goal here was to add unit testing around the fetch items use case (now `FetchArtistItemsUseCase` so I could easily drive out more advanced logic for some of the more fiddly situations we can run into (two artists matched across a single provider for example). Additionally, this would let me drop some of our existing, more awkward, tests in `ArtistTest` to that level. To do that, there were two things I needed to do to lower the pain threshold:

- Extracting an interface for `MediaItemRepository` to allow it to be easily faked/stubbed.
- Creating a dedicated `ViewModel` for the artist details screen to detangle the artist specific parts from the more generalized code in `ItemDetailsViewModel`.

Those are both done, but the rabbit hole was realizing I needed to duplicate the behaviour of keeping items on screen up to date with the server (favourites being the easiest way to demonstrate that not working).


composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/artist/ArtistDetailsViewModel.kt
```diff
@@ -0,0 +1,200 @@
...
+                    .take(ARTIST_SECTION_LIMIT),
+                itemList = ItemList.ArtistLibrary(artist.itemId),
+            )
+        }
+        .stateIn(viewModelScope, SharingStarted.Eagerly, DataState.Loading())
+
+    private val allItems = MutableStateFlow<DataState<List<AppMediaItem>>>(DataState.Loading())
+    private val allProviderFilter = MutableStateFlow<Section.ProviderFilter?>(null)
+    val all = allItems
+        .updateFrom(mediaItemRepository)
```
seadowg: `updateFrom` encourages using multiple `Flow` objects as opposed to a single state container in a ViewModel (or breaking ViewModels up into smaller ones), but I don't think that's a problem: large state containers mutated through `update`/`copy` can start to get unwieldy pretty quickly, and it's easy to make a mistake with what gets overriden where.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/artist/ArtistDetailsViewModel.kt
```diff
@@ -0,0 +1,200 @@
...
+import io.music_assistant.client.ui.compose.item.ItemDetailsViewModel.Companion.ARTIST_SECTION_LIMIT
+import io.music_assistant.client.ui.compose.item.ItemList
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+class ArtistDetailsViewModel(
```
seadowg, line 25: There's a bunch of duplication in this file, but I'd like to get feedback on `updateFrom` before I clean it up.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/FetchArtistItemsUseCase.kt
```diff
@@ -6,12 +6,11 @@ import io.music_assistant.client.data.model.client.items.Artist
 import io.music_assistant.client.data.model.server.ProviderMapping
 import io.music_assistant.client.data.repository.MediaItemRepository
 
-object ItemUseCases {
-    suspend inline fun <reified T : AppMediaItem> fetchArtistItemsAcrossProviders(
-        mediaItemRepository: MediaItemRepository,
+class FetchArtistItemsUseCase(private val mediaItemRepository: MediaItemRepository) {
```
seadowg, line 9: For actual prod code organization, I really like having use cases in domain area objects (like the previous `ItemUseCases` setup), but it gets really annoying for organizing tests: IDEs pretty much all prefer that the test name and the class name match and reward you for it (CMD + Shift + T in Jetbrains products for example).

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -198,6 +151,82 @@ class MediaItemRepository(
...
+    /**
+     * Create a nullable default version of [MediaItemRepository.itemChanges] so that the result of
+     * `combine` doesn't wait for a change to emit. There isn't currently a [Flow] transformation
+     * that behaves like this: `zip` emits new pairs (it "buffers" one side until the other emits
+     * a corresponding value) and `combine` only emits once both [Flow] instances have emitted.
+     */
+    val nullableItemChanges: Flow<MediaItemChange?> = mediaItemRepository.itemChanges.map { it }
+    val itemChangesWithDefault = nullableItemChanges.onStart { emit(null) }
+
+    return this.combine(itemChangesWithDefault) { data, change ->
```
formatBCE: `combine` applies only the latest MediaItemChange to the original fetched list.
That means, every intermittent changes will be lost, as far as I see...

I mean, for this to work item changes should be cumulative, probably. Instead of keeping cumulative changes in the state flow, we'd need to hold it somewhere else.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/artist/ArtistDetailsViewModel.kt
formatBCE: I think we're digging a hole with this approach. We either miss consecutive updates - or, in case we will make list and gonna hold whole history - we would have missed actual server-state change at some point, not to mention "heavy" applications of every change on every update.

formatBCE (commented):
Overall it's good start - replacing single god-like entity with specific models. I love it, and always prefer this approach. However, we need to think on ultimate solution of merging server loaded entities, server events, and locally-applied changes like favs - to single item list. Priority, consecutive changes handling, integrity.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
seadowg: Ach, that seems obvious now after coming back to it! Clearly got too deep into the weeds of trying to make this fit existing transformations. I'll have another thing - if there isn't a way to have it work elegantly as part of a transformation chain, the backup is just to use a convenience update method for the root `MutableStateFlow`.

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/common/DataState.kt
```diff
@@ -44,3 +46,7 @@ fun <T, U> DataState<T>.map(transform: (T) -> U): DataState<U> {
         )
     }
 }
+
+fun <T, U> Flow<DataState<T>>.mapData(transform: (T) -> U): Flow<DataState<U>> {
```
formatBCE, line 50: Nice!

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
```diff
@@ -0,0 +1,111 @@
...
+                stateFlow.value =
+                    DataState.Data(mediaItemRepository.fetchMediaItems(it).getOrEmptyList())
+            } catch (_: Exception) {
+                stateFlow.value = DataState.Error()
+            }
+        }
+    }
+}
+
+private fun <T : AppMediaItem> List<T>.replacing(changed: T): List<T> =
```
seadowg, line 110: I think we might need to make `replacing` more complex in the future: `hasMapping` lets us know that we're referring to the same "thing" (however it's mapping), but we might only want to update a subset of properties based on that. An `Updated` event for the Tidal version of a track probably shouldn't affect the Spotify version's `provider` for example.

That's probably something to look at down the road if we actually see that causing problems in the wild though.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
```diff
@@ -0,0 +1,111 @@
...
+    private val mediaItemRepository: MediaItemRepository,
+) {
+    private val stateFlow = MutableStateFlow(initial)
+    private var request: Request? = null
+
+    /**
+     * Retrieve and store items using [request]. [request] will also be used to update the items
+     * if/when needed.
+     */
+    suspend fun set(request: Request) {
```
formatBCE, line 37: Maybe `init`? It's better than `set` IMHO...

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
formatBCE: Yeah we could add lambda parameter that will receive old T and return new T - and by default it would return changes one.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -198,8 +146,53 @@ class MediaItemRepository(
     }
 }
 
-private fun <T : AppMediaItem> List<T>.replacing(changed: T): List<T> =
-    map { if (it.itemId == changed.itemId) changed else it }
+/**
+ * The home-page recommendation rows as the server returned them, with or
+ * without embedded items (see [supportsRecommendationRowItems]).
+ */
+suspend fun MediaItemRepository.fetchRecommendationRows(): Result<List<RecommendationFolder>> =
```
formatBCE, line 153: Should this be extension function on any MediaItemRepository though?

formatBCE (approved):
Overall it's good. Couple minor questions, but might be follow-up.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
seadowg: I think we probably want the recommendation stuff converted to a use, but I was mainly just trying to cut down the `MediaItemRepository` surface here. OK to leave for now?

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
seadowg: Yeah, I've had a hard time with the naming here. You can call `set `multiple times though as well so `init` (which I'd also considered) feels a little misleading. `load`?

formatBCE: 'activate'?  :)

seadowg: Again that still implies just a one-time operation I think. Let's stick with `set` right now? It's an easy rename (thanks JetBrains) once we think of something better.

formatBCE: Okay... :)

PR #963, opened 2026-09-03
Prevent WebRTC sends from crashing during close

teancom:

Finally got the crash logs to download from the OHF org and immdiately spotted these from field crash reports.


composeApp/src/commonMain/kotlin/io/music_assistant/client/player/sendspin/transport/WebRTCDataChannelTransport.kt
```diff
@@ -79,8 +79,11 @@ class WebRTCDataChannelTransport(
     override suspend fun sendText(message: String) {
         val currentState = dataChannelWrapper.state.value
         if (currentState != DataChannelState.Open) {
-            logger.w { "Attempted to send text while channel not open (state: $currentState)" }
-            error("Channel not open (state: $currentState)")
```
formatBCE, line 83: You removed error here - this means we don't send, but also don't disconnect. But the channel is closed, we should disconnect, aren't we?

formatBCE (approved):

PR #973, opened 2026-09-05
Handle the iOS Local Network permission on first server connect

### LINKED ISSUE #956: New setup initially throws exception
## Context
 - Operating system and version: iOS 26.6.1
 - App version: 1.0 (Build 49)
 - Music Assistant server version: 2.10.1
 - Music Assistant server installation method (Home Assistant App or Docker image): Docker

### Problem description

When setting up the app from fresh installation and connecting to the MA server for the first time, an error is displayed (see below). I believe this is because permissions have not yet been granted for the app to be able to find devices on the local network - this capability is offered by iOS AFTER the intial connection is attempted and fails. After that, connection is fine.
```
Exception in http request: Error Domain=NSURLErrorDomain Code=1009 "The internet connection appears to be offline."
```
### Steps to reproduce

1. [Remove app, then] Download app
2. In Settings, enter server and port
3. Click connect

### Expected behavior

Server should connect without error. If allowing connectivity from the app to other devices on the network is a pre-req, the offer to do so should precede the ability to connect to the MA server.

teancom:
Closes #956 

Prompt the user on a fresh install to allow Local Networking access to the app and connect without an error or having them tap 'connect' again if they do so. Give direct advice on how to fix it if they accidentally click the wrong button. Only for iOS at the moment, though once we move to SDK 37 / Android 17 apparently we'll have to do it there too, so I plumbed it as generically as I could.

iOS doesn't provide a way to just read if we have local networking support or not. Apparently the "recommended" approach is what I took, which is to do a 'fake' mDNS query across loopback. Annoying, but I did a bunch of testing on my actual phone and it all works.


formatBCE (commented):
I don't like this... Too much iOS specific code is slipping through into common code. Strings containing iOS specific data don't have iOS in IDs...

Can you make it cleaner @teancom please?

formatBCE (approved):

PR #975, opened 2026-09-06
Rewrite the Sendspin local player as a standalone :sendspin module

### LINKED ISSUE #978: iOS: system Play after background pause tears down reconnect and waits for foreground
## Context
- Operating system and version: iOS 26.6.1.
- App version: 1.0 (50).
- Music Assistant server version: 2.10.2.
- Music Assistant server installation method: Home Assistant OS (HAOS), Docker container.

### Problem description

Resuming the local Sendspin player from system media controls can fail after pausing in the background. The app receives Play, but a stale WebSocket triggers background teardown while Play is pending. Opening the app restores the session; in one captured attempt the queued Play then starts audio without another press.

A short pause of approximately 2.3 seconds resumed successfully. Failures were observed after approximately 10 seconds and after longer pauses, with both FLAC and PCM.

### Steps to reproduce

1. Play through the local iOS player and background the app.
2. Pause using system media controls.
3. Wait long enough for the paused connection to become stale, then press Play without opening the app.
4. Playback stays stopped. Open the app and observe connection recovery.

### Expected behavior

An explicit system Play should recover the authenticated MA session and Sendspin connection without requiring foreground UI. A subsequent Pause should cancel pending Play, including replay after a later foreground.

Sanitized sequence:

    NativeAudioController: Remote command: play
    LocalPlayerCtrl: Remote command: play
    NativeAudioController: resumeSink
    DirectTransport: ... NSPOSIXErrorDomain Code=53 "Software caused connection abort"
    ServiceClient: Transport→Reconnecting while backgrounded → Disconnected.Backgrounded
    SendspinWsHandler: Disconnecting WebSocket (explicit)
    SendspinClient: Closing Sendspin client
    LocalPlayerCtrl: Local command with no live Sendspin transport — reviving and queueing
    LocalPlayerCtrl: Cannot create Sendspin client: MA session not established — deferring Sendspin start
    LocalPlayerCtrl: Pending play timed out without server confirmation; clearing spinner
    ... app opened ...
    ServiceClient: Force reconnect: was Backgrounded
    AuthMgr: Authenticated
    LocalPlayerCtrl: Draining 1 queued commands
    NativeAudioController: AudioQueue started

The log does not establish why the socket was aborted. It does show that Play reaches the app and recovery is then disabled before playback becomes active. AirPods removal also causes pauses, but a corresponding return-to-ear Play was not captured; this report does not assume all AirPods auto-resume failures share this cause.

I have a candidate fix and regression tests prepared: a bounded playback-request lease, session recovery before Sendspin registration, and cancellation on Pause/reset/timeout. May I work on this issue and submit the candidate as a draft PR for discussion and device testing?

### LINKED ISSUE #982: not played Android
## Context
 - Operating system and version: haos 18.2
 - App version: 0.13 android
 - Music Assistant server version: 2.10.2
 - Music Assistant server installation method (Home Assistant App or Docker image): Home Assistant App

### Problem description

1788868446402 Info/ServiceClient: isReadyForCommands=false
1788868446405 Info/ServiceClient: Network became available while not fully connected — rotating HttpClient
1788868446412 Info/ServiceClient: Playback inactive (state=Disconnected.Initial)
1788868446414 Info/MainDataSource: SessionState changed: je4
1788868446414 Info/MainDataSource: Disconnected (je4) - clearing data
1788868446414 Info/SendspinClientFactory: Destroying shared audio pipeline
1788868446414 Info/MainDataSource: Clearing all cached data
1788868446527 Info/ServiceClient: App foregrounded (state=Disconnected.Initial)
1788868446616 Info/MainDataSource: SessionState changed: fe4
1788868446616 Info/MainDataSource: Connecting - stopping Sendspin
1788868446774 Info/MainDataSource: SessionState changed: ce4
1788868446774 Info/ServiceClient: Transport→Connected (wasReconnecting=false, needsReauthOnReconnect=true, dcs=AwaitingServerInfo)
1788868446774 Info/AuthMgr: AwaitingServerInfo
1788868446774 Info/MainDataSource: Clearing all cached data
1788868446789 Info/MainDataSource: SessionState changed: ce4
1788868446789 Info/MainDataSource: Clearing all cached data
1788868446822 Info/AuthMgr: AwaitingAuth(NotStarted) — auto-login with saved token
1788868446823 Info/MainDataSource: SessionState changed: ce4
1788868446823 Info/MainDataSource: Clearing all cached data
1788868446903 Info/MainDataSource: SessionState changed: ce4
1788868446905 Info/AuthMgr: Authenticated
1788868446906 Info/ServiceClient: isReadyForCommands=true
1788868446922 Info/SendspinClientFactory: Creating Sendspin client over WebSocket (proxy mode, shared pipeline)
1788868446936 Error/SendspinClientFactory: Failed to create and start Sendspin client
java.security.NoSuchAlgorithmException: X25519 KeyPairGenerator not available
	at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:253)
	at a12.invoke(Unknown Source:17)
	at sl3.a(Unknown Source:20)
	at hs.j(Unknown Source:121)
	at xc4.a(Unknown Source:96)
	at x1.invokeSuspend(Unknown Source:239)
	at ut.resumeWith(Unknown Source:7)
	at vu0.run(Unknown Source:113)
	at jl0.run(Unknown Source:114)

1788868446936 Warn/LocalPlayerCtrl: Cannot create Sendspin client: X25519 KeyPairGenerator not available
1788868446937 Info/MainDataSource: Updating players and queues
1788868446941 Info/ServiceClient: #a026a202 player_queues/all command sent
1788868447058 Info/LocalPlayerCtrl: Injecting synthetic local player
1788868447065 Info/ServiceClient: #a026a202 player_queues/all command acked
1788868447123 Info/ServiceClient: #0d883f33 player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command sent
1788868447123 Info/ServiceClient: #7bd3c9a4 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command sent
1788868447123 Info/ServiceClient: #c3b79d5f player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command sent
1788868447164 Info/ServiceClient: #0d883f33 player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command acked
1788868447164 Info/ServiceClient: #c3b79d5f player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command acked
1788868447166 Info/ServiceClient: #7bd3c9a4 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command acked
1788868449580 Info/ServiceClient: #14368b9a player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command sent
1788868449590 Info/ServiceClient: #14368b9a player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command acked
1788868450424 Info/ServiceClient: #3550be45 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command sent
1788868450443 Info/ServiceClient: #3550be45 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command acked
1788868451076 Info/ServiceClient: #5a862b5c player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command sent
1788868451106 Info/ServiceClient: #5a862b5c player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command acked
1788868451903 Info/ServiceClient: #76490898 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command sent
1788868451920 Info/ServiceClient: #76490898 player_queues/items target=media_player.lg_webos_tv_lf652v_2 command acked
1788868452256 Info/ServiceClient: #ca5350cc player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command sent
1788868452277 Info/ServiceClient: #ca5350cc player_queues/items target=media_player.malenkaia_komnata_home_assistant_voice_speaker_media_player command acked
1788868460624 Info/PlayDispatch: HomeScreenViewModel: uri=library://track/1310 option=REPLACE radio=false queue=[REDACTED_ID]
1788868460625 Info/ServiceClient: #db794edc player_queues/play_media target=[REDACTED_ID] command sent
1788868460653 Error/RpcEngine: RPC error for message [REDACTED_ID]: Answer(json={"message_id":"[REDACTED_ID]","error_code":10,"details":"Плеер недоступен."})
1788868460657 Info/ServiceClient: #db794edc player_queues/play_media target=[REDACTED_ID] command acked
1788868464200 Info/SendspinClientFactory: Destroying shared audio pipeline
1788868464201 Info/AudioStreamManager: Stopping stream
1788868464201 Info/AudioStreamManager: Closing AudioStreamManager
1788868464201 Info/MediaPlayerController: Stopping raw PCM stream
1788868482520 Info/SendspinClientFactory: Creating Sendspin client over WebSocket (proxy mode, shared pipeline)
1788868482521 Error/SendspinClientFactory: Failed to create and start Sendspin client
java.security.NoSuchAlgorithmException: X25519 KeyPairGenerator not available
	at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:253)
	at a12.invoke(Unknown Source:17)
	at sl3.a(Unknown Source:20)
	at hs.j(Unknown Source:121)
	at xc4.a(Unknown Source:96)
	at x1.invokeSuspend(Unknown Source:239)
	at ut.resumeWith(Unknown Source:7)
	at vu0.run(Unknown Source:113)
	at jl0.run(Unknown Source:114)

1788868482521 Warn/LocalPlayerCtrl: Cannot create Sendspin client: X25519 KeyPairGenerator not available
1788868482521 Info/LocalPlayerCtrl: Injecting synthetic local player
1788868490447 Info/PlayDispatch: HomeScreenViewModel: uri=library://track/1310 option=REPLACE radio=false queue=[REDACTED_ID]
1788868490448 Info/ServiceClient: #67bddb5c player_queues/play_media target=[REDACTED_ID] command sent
1788868490479 Error/RpcEngine: RPC error for message [REDACTED_ID]: Answer(json={"message_id":"[REDACTED_ID]","error_code":10,"details":"Плеер недоступен."})
1788868490493 Info/ServiceClient: #67bddb5c player_queues/play_media target=[REDACTED_ID] command acked

### Steps to reproduce

The steps that we can take to reproduce your issue. Ideally format it as a list like:

1. The caveat is that the connection is made via the SSL protocol to the external domain address, not the internal server address.
2. The latest version that works properly is 0.10 
 log on 0.10
1788870058138 Info/ServiceClient: isReadyForCommands=false
1788870058141 Info/ServiceClient: Network became available while not fully connected — rotating HttpClient
1788870058145 Info/MainDataSource: SessionState changed: u24
1788870058146 Info/MainDataSource: Disconnected (u24) - clearing data
1788870058146 Info/SendspinClientFactory: Destroying shared audio pipeline
1788870058146 Info/MainDataSource: Clearing all cached data
1788870058149 Info/ServiceClient: Playback inactive (state=Disconnected.Initial)
1788870058253 Info/ServiceClient: App foregrounded (state=Disconnected.Initial)
1788870058645 Info/MainDataSource: SessionState changed: q24
1788870058645 Info/MainDataSource: Connecting - stopping Sendspin
1788870058755 Info/ServiceClient: Transport→Connected (wasReconnecting=false, needsReauthOnReconnect=true, dcs=AwaitingServerInfo)
1788870058755 Info/MainDataSource: SessionState changed: n24
1788870058756 Info/AuthMgr: AwaitingServerInfo
1788870058756 Info/MainDataSource: Clearing all cached data
1788870058762 Info/MainDataSource: SessionState changed: n24
1788870058762 Info/MainDataSource: Clearing all cached data
1788870058800 Info/AuthMgr: AwaitingAuth(NotStarted) — auto-login with saved token
1788870058801 Info/MainDataSource: SessionState changed: n24
1788870058801 Info/MainDataSource: Clearing all cached data
1788870059001 Info/MainDataSource: SessionState changed: n24
1788870059001 Info/ServiceClient: isReadyForCommands=true
1788870059003 Info/SendspinClientFactory: Creating Sendspin client over WebSocket (proxy mode, shared pipeline)
1788870059003 Info/AuthMgr: Authenticated
1788870059008 Info/SendspinClient: Starting Sendspin client: mix 3
1788870059009 Info/SendspinClient: Connecting to Sendspin with transport
1788870059010 Info/SendspinClient: Initializing with system volume: 0%
1788870059013 Info/MessageDispatcher: Starting MessageDispatcher
1788870059013 Info/SendspinWsHandler: Connecting to server
1788870059033 Info/SendspinWsHandler: Connected to server (epoch 1)
1788870059034 Info/SendspinClientFactory: Sendspin client started via WebSocket
1788870059035 Info/LocalPlayerCtrl: Sendspin client initialized and started
1788870059035 Info/MainDataSource: Updating players and queues
1788870059037 Info/ServiceClient: #a7215849 player_queues/all command sent
1788870059039 Info/LegacySession: Received auth_ok - authentication successful
1788870059050 Info/SendspinClient: Protocol ready (server=Music Assistant (d5369777-music-assistant), trust=NONE)
1788870059050 Info/SendspinClient: Protocol established — transitioning to Ready
1788870059051 Info/ServiceClient: #a7215849 player_queues/all command acked
1788870059051 Info/MessageDispatcher: Received server/hello from Music Assistant (d5369777-music-assistant)
1788870059051 Info/MessageDispatcher: Sending client/state: SYNCHRONIZED
1788870059055 Info/LocalPlayerCtrl: Injecting synthetic local player
1788870059072 Info/ClockSync: Clock sync quality: LOST -> GOOD (rtt=18ms)
1788870059127 Info/ServiceClient: #c48e9e3b player_queues/items target=media_player.lg_webos_tv_lf652v_2 command sent
1788870059127 Info/ServiceClient: #077af959 player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command sent
1788870059135 Info/ServiceClient: #077af959 player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command acked
1788870059135 Info/ServiceClient: #c48e9e3b player_queues/items target=media_player.lg_webos_tv_lf652v_2 command acked
1788870059474 Info/ServiceClient: #c2f12a92 player_queues/items target=[REDACTED_ID] command sent
1788870059562 Info/ServiceClient: #c2f12a92 player_queues/items target=[REDACTED_ID] command acked
1788870060052 Info/MainDataSource: Updating players and queues
1788870060052 Info/ServiceClient: #767d3e0c player_queues/all command sent
1788870060052 Info/ServiceClient: #674c664e player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command sent
1788870060053 Info/ServiceClient: #278146bc player_queues/items target=media_player.lg_webos_tv_lf652v_2 command sent
1788870060054 Info/ServiceClient: #a7376204 player_queues/items target=[REDACTED_ID] command sent
1788870060074 Info/ServiceClient: #674c664e player_queues/items target=ma_companion_35eae046-a841-44b1-8879-65bd4ed25f5c command acked
1788870060095 Info/ServiceClient: #278146bc player_queues/items target=media_player.lg_webos_tv_lf652v_2 command acked
1788870060103 Info/ServiceClient: #767d3e0c player_queues/all command acked
1788870060188 Info/ServiceClient: #a7376204 player_queues/items target=[REDACTED_ID] command acked

formatBCE:
Replace the early-adoption Sendspin stack with a pure KMP module that has one reconnect policy, structured concurrency per connection attempt, bounded inbound channels, a byte-capped jitter buffer, a seeded Kalman clock over probe bursts (also the liveness watchdog), and a scheduler that corrects against sink head position. Only the Noise-encrypted session is supported; the plaintext legacy session is dropped.

App side: LocalPlayerAdapter keeps optimistic UI, the offline command queue, and reconciliation; LocalPlayerEndpoints derives the endpoint from the MA session and keeps it through transient session states; platform sinks and decoders move to player/local. Remove LocalPlayerController, MediaPlayerController, the old player/sendspin package, and the require-encryption and client-id settings.

Fixes #978 
Fixes #982


sendspin/src/commonMain/kotlin/io/music_assistant/sendspin/audio/Scheduler.kt
```diff
@@ -0,0 +1,284 @@
...
+                    SinkEvent.RouteChanged -> logger.i { "Audio route changed" }
+                }
+                pipeline.wakeups.trySend(Unit)
+            }
+        }
+        publish(AudioPhase.Buffering, starved = false)
+    }
+
+    /** Returns true when [chunk] was consumed (played or dropped); false to retry later. */
+    private suspend fun play(chunk: AudioChunk): Boolean {
```
teancom, line 169: `step()` peeks at a packet and removes it only when `play()` returns `true`. But `play()` calls `decoder.decode(chunk)` before deciding whether to wait for clock synchronization or the packet’s scheduled playback time.

When it waits and returns `false`, the next iteration decodes the same packet again. New incoming packets can wake the scheduler early, repeating this multiple times.

That is not safe for Android’s stateful codecs:

• Opus advances the same Concentus decoder on every invocation.
• FLAC submits the input to MediaCodec again.

The earlier decoded output is discarded; the eventually played output comes from an incorrectly advanced decoder state. This can corrupt audio during ordinary buffering/timing waits. The old implementation gated timing before decoding.

Fix: retain decoded output across retries, or structure scheduling so each encoded packet is decoded exactly once. Add a counting/stateful fake-decoder test for a future-dated packet and an initially unavailable clock.

sendspin/src/commonMain/kotlin/io/music_assistant/sendspin/audio/Scheduler.kt
```diff
@@ -0,0 +1,284 @@
...
+            error("unreachable")
+        } finally {
+            closeSink()
+            decoder.close()
+        }
+    }
+
+    private suspend fun step(scope: kotlinx.coroutines.CoroutineScope) {
+        val state = pipeline.stream.value
+        if (state.flushGeneration != seenFlush) {
```
teancom, line 78: The clear path flushes the sink and resets timing counters, but does not reset `played`. After audio has played once, subsequent writes following `stream/clear` therefore never emit another `AudioEvent.Started`.

This matters beyond telemetry:

1. A local seek freezes the position tracker pending actual audio confirmation.
2. `stream/clear` drops the old buffered audio.
3. Audio resumes without a fresh `stream/start`.
4. No `PlaybackStarted` event reaches `LocalPlayerAdapter`, so its `positionTracker.confirmPlaying()` call never happens.

The position can remain frozen despite resumed playback. A fresh `stream/start` masks the bug by rebuilding the scheduler state, but the clear-and-continue path is explicitly supported.

Fix: reset playback-confirmation state on clear and test play → clear → more audio, asserting a second `Started` event.

teancom (changes requested):
I haven't loaded this on my phone for field testing yet, but Astra had some words for Fable.

sendspin/src/commonMain/kotlin/io/music_assistant/sendspin/audio/Scheduler.kt
```diff
@@ -0,0 +1,302 @@
...
+     * Returns true when [chunk] was consumed (played or dropped); false to retry later.
+     *
+     * Every "retry later" exit happens before the decode: decoders are stateful
+     * (Opus advances per call, MediaCodec queues the frame), so a chunk is decoded
+     * exactly once, on the iteration that consumes it.
+     */
+    private suspend fun play(chunk: AudioChunk): Boolean {
+        val out = handle ?: return false
+        val fmt = format ?: return false
+        val target = clockSync.toLocalMicros(chunk.timestampMicros)?.plus(pipeline.userDelayMicros)
```
teancom, line 172: The old scheduler calculated:
target = serverTimeToLocal(timestamp) - userDelayMicros

The new scheduler calculates:
target = clockSync.toLocalMicros(timestamp) + pipeline.userDelayMicros

The persisted setting and its mapping are unchanged: `sendspinStaticDelayMs` flows into `userDelayMs`, then is multiplied by 1,000 without negation. `SettingsRepository` still documents positive values as playing earlier to compensate for downstream audio latency, with a default of +250 ms.

User-visible consequence: a user’s existing latency compensation now moves playback in the opposite direction. With the default setting, the scheduling target changes from 250 ms before the timestamp to 250 ms after it—a 500 ms shift relative to the previous target, before accounting for the new sink-feedback behavior. This undermines synchronization with other players, especially on the explicitly open-loop iOS path.

Suggested fix: preserve the existing setting’s semantics by subtracting the delay. If the module intentionally defines positive delay as “play later,” negate the value at the app-to-module boundary instead.

Regression test: verify that a persisted positive compensation value advances the target, and a negative value delays it. The existing test checks the new addition behavior, so it does not catch compatibility with the old setting.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalPlayerAdapter.kt
```diff
@@ -0,0 +1,589 @@
...
+                StopCause.SinkFailed -> pauseLocalIfPlaying()
+                StopCause.ServerEnded, StopCause.Cleared, StopCause.Disabled -> Unit
+            }
+
+            PlayerEvent.FocusRegained -> if (pausedByInterruption) {
+                pausedByInterruption = false
+                localPlayerData.value?.let { handleLocalCommand(it, PlayerAction.Play) }
+            }
+
+            PlayerEvent.ServerRefreshNeeded -> {
```
teancom: The adapter still queues local commands when the MA command transport is unavailable. However, the only remaining trigger for `drainCommandQueue()` is:
PlayerEvent.ServerRefreshNeeded -> {
    _needsServerRefresh.emit(Unit)
    drainCommandQueue()
}

That event comes from Sendspin activation, not MA control-connection recovery. The old `MainDataSource` reconnect handlers explicitly drained the queue; those calls were removed.

Reproduction sequence:

1. Local playback is active.
2. The MA control WebSocket disconnects, but the separate Sendspin connection survives.
3. The user taps Pause, Play, or Seek while `isReadyForCommands` is false; the request is queued.
4. The MA control connection recovers with the same endpoint/token.
5. Sendspin does not reconnect or reactivate, so nothing drains the command queue.

User-visible consequence: the requested pause/play/seek is not applied when connectivity returns. It remains queued until an unrelated future Sendspin activation, or is discarded when local state is cleared. For Pause, the optimistic UI may briefly show paused before a server refresh restores the still-playing state.

This is specifically the MA-only outage case. When both connections drop, Sendspin’s subsequent activation can mask the bug by draining the queue. Keeping Sendspin alive through transient MA disconnections is an explicit goal of the rewrite, so this is a supported scenario rather than an arbitrary edge case.

Suggested fix: restore a drain trigger when the MA command transport becomes ready, independently of Sendspin activation. Coordinate it with the existing activation-triggered drain.

Regression test: keep Sendspin connected, make the command transport unavailable, queue a command, restore command readiness without emitting another activation, and assert that the command is sent exactly once.

(Back to my voice) I would say this is *precisely* the sort of thing that drove this rewrite in the first place and I would *not* immediately jump to "replicate how we did it before", but yeah. Addressing this sort of edge-case is the hard bit. :-/

teancom (changes requested):
Round two. Previous findings were addressed to Astra's satisfaction. (And mine).

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalPlayerEndpoints.kt
```diff
@@ -0,0 +1,153 @@
...
+        settings.sendspinUseCustomConnection,
+        settings.sendspinHost,
+        settings.sendspinPort,
+        settings.sendspinPath,
+        settings.sendspinUseTls,
+    ) { values -> values }
+        .runningFold(null as Endpoint?) { previous, values -> next(previous, values[0] as SessionState) }
+        .stateIn(scope, SharingStarted.Eagerly, null)
+
+    private fun next(previous: Endpoint?, state: SessionState): Endpoint? = when (state) {
```
teancom: The endpoint reducer uses:
is SessionState.Connected -> connected(previous, state) ?: previous

But `connected()` deliberately returns `null` for terminal authentication states:
if (connection is DataConnectionState.AwaitingAuth &&
    (connection.authProcessState is AuthProcessState.LoggedOut ||
     connection.authProcessState is AuthProcessState.Failed)
) {
    return null
}

The caller converts that intentional “clear the endpoint” result into “keep the previous endpoint.”

Trigger: an established MA session fails reauthorization and remains in `Connected` with `AwaitingAuth(Failed)`. The API implementation explicitly allows this state to persist until manual login or another transport bounce.

Consequence: the UI clears its data, but the independent Sendspin player remains enabled with its previous endpoint. Existing playback or reconnect attempts can continue when terminal auth failure was supposed to stop them. This is a lifecycle/teardown defect—not a demonstrated authentication bypass.

Why introduced: the old `MainDataSource` explicitly stopped the local player on terminal auth failure. The rewrite removes that call and delegates ownership to this endpoint flow, whose null fallback prevents teardown.

Suggested fix: distinguish keep, replace, and clear outcomes, or handle terminal auth states before applying the fallback.

Regression test: start with a valid endpoint, transition to each of `AwaitingAuth(Failed)` and `AwaitingAuth(LoggedOut)`, and assert that the endpoint becomes null and the player is disabled. Keep a separate test proving transient auth states retain the endpoint.

sendspin/src/commonMain/kotlin/io/music_assistant/sendspin/audio/Scheduler.kt
```diff
@@ -0,0 +1,302 @@
...
+        }
+        handle = opened
+        format = pcmFormat
+        corrector = if (pcmFormat.isPcm) DriftCorrector(pcmFormat.channels, pcmFormat.bitDepth / BITS_PER_BYTE) else null
+        openGeneration = state.generation
+        framesWritten = 0
+        played = false
+        lateDrops = 0
+        insertedSilenceMicros = 0
+        scope.launch {
```
teancom: Every sink replacement launches a collector into the scheduler’s long-lived scope:
scope.launch {
    opened.events.collect { event ->
        // Applies events to the current pipeline.
    }
}

`closeSink()` closes the handle but does not cancel its collector. Both platform implementations expose a `MutableSharedFlow`, which does not complete when the handle closes.

Confirmed consequence: each stream replacement leaves another suspended collector alive until the entire scheduler is cancelled. Collector lifetime is incorrectly tied to the enabled player rather than to the sink handle.

Additional consequence, timing-dependent: a focus-loss or failure event already buffered for handle A can be delivered after handle B replaces it. The collector does not check handle identity or generation; it calls `pipeline.onSinkFailure()`, which ends and clears the current stream. A stale callback can therefore stop replacement playback.

Suggested fix: retain and cancel the per-handle collector job during replacement/close, and guard event handling against stale handle identity or generation.

Regression tests:

• Repeatedly replace sinks and assert old event subscriptions are removed.
• Arrange a pending failure event from A, replace it with B, then deliver the event and assert B remains active.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/LocalCommandQueue.kt
```diff
@@ -0,0 +1,93 @@
...
+    private val mutex = Mutex()
+    private val queue = mutableListOf<Entry>()
+
+    private data class Entry(val action: PlayerAction, val request: Request)
+
+    init {
+        scope.launch { isReady.filter { it }.collect { drain() } }
+    }
+
+    suspend fun sendOrQueue(action: PlayerAction, request: Request) {
```
teancom, line 50: ### Command-queue draining: right architecture, one remaining race ###

The new `LocalCommandQueue` observes MA command readiness directly, and the Sendspin-activation drain is removed. This is better than restoring scattered reconnect hooks.

The added test covers the original scenario: queue a command during an MA-only outage, restore readiness without Sendspin activation, and send it once. That reported sequence is fixed.

However, `LocalCommandQueue.kt:36–41` can enqueue after the readiness-triggered drain has already completed:

1. A direct `send(request)` is in flight.
2. The connection drops and recovers while that request is pending.
3. Readiness becomes true; the collector drains an empty queue.
4. The pending send returns failure, and `sendOrQueue()` enqueues it.
5. Readiness remains true, so there is no further drain trigger.

The request is stranded until another readiness transition—the same symptom through a narrower timing window.

Follow-up: serialize queue admission and readiness handling, or ensure enqueueing while ready wakes a single drain worker. Add a suspended-send test that restores readiness before completing the send with failure, then asserts replay without another readiness change. Avoid an immediate unbounded retry loop for persistent failures.

teancom (changes requested):
This is my last round for the night, but this is looking better and better. Thank you!!

sendspin/src/commonMain/kotlin/io/music_assistant/sendspin/transport/TransportConnector.kt
```diff
@@ -0,0 +1,44 @@
...
+                throw e
+            }
+        }
+    }
+
+    companion object {
+        private val PING_INTERVAL = 5.seconds
+
+        /** Production connector: WebSockets over a client derived from the app's. */
+        fun ktor(httpClient: HttpClient): TransportConnector {
```
teancom, line 38: Each enabled player lifetime creates a connector, which lazily creates a derived client:
val wsClient by lazy {
    httpClient.config {
        install(WebSockets) { /* … */ }
    }
}

Neither the connector nor the enabled-lifetime teardown closes that client. Closing its WebSocket session is insufficient.

Trigger: repeatedly disable and re-enable Sendspin, allowing a WebSocket connection during each enabled lifetime.

Impact: unclosed client resources and engine references accumulate. Ktor 3.5.2’s derived clients share the source engine, but have their own client jobs and, for managed engines, their own engine-reference-count entries.

This is not a new engine per reconnect: ordinary reconnects within one enabled lifetime reuse the derived client.

Why introduced: the old WebSocket handler explicitly closed its owned client.

Suggested fix: give the derived client an explicit owner and close it when that enabled lifetime ends. Do not close the app-owned source client.

Regression test: repeat enable → connect → disable and verify each derived client is closed while the source client remains usable.

This also applies to  `SendspinPlayerImpl.kt:87–104`

composeApp/src/androidMain/kotlin/io/music_assistant/client/player/local/AudioTrackSink.kt
```diff
@@ -0,0 +1,216 @@
...
+
+        override fun resume() {
+            runCatching { track.play() }
+        }
+
+        override fun flush() {
+            runCatching { track.flush() }
+        }
+
+        override fun position(): SinkPosition? {
```
teancom: The sink returns a wrapping 32-bit playback position. The scheduler compares it with a continuously increasing `Long` count of written frames:
framesWritten - playedByNow

There is no wrap-extension logic.

Trigger: one uninterrupted sink lifetime exceeds 2³² frames, without a flush or rebuild:

• Approximately 24.86 hours at 48 kHz
• Approximately 6.21 hours at 192 kHz

Impact: when the device counter wraps, the scheduler suddenly calculates roughly one counter-cycle of nonexistent queued audio. Incoming chunks appear extremely late and are dropped, producing sustained silence rather than normal playback.

Why introduced: the old scheduler did not use device-position feedback, so it did not mix these counter bases.

Suggested fix: extend the wrapping device counter into a monotonic frame count and reset that tracking consistently on flush/rebuild. Handle transitions between timestamp and fallback position sources consistently.

Regression test: feed positions immediately before and after wrap and assert that queued duration stays small and continuous.

Related:  `Scheduler.kt:253–255`

composeApp/src/iosMain/kotlin/io/music_assistant/client/player/local/AudioQueueSink.kt
```diff
@@ -0,0 +1,118 @@
...
+ */
+class AudioQueueSink(private val onRemoteCommand: (command: String) -> Unit) : AudioSink {
+    private val logger = Logger.withTag("AudioQueueSink")
+
+    override fun open(format: SinkFormat): SinkHandle {
+        val player = PlatformPlayerProvider.player ?: error("no PlatformAudioPlayer registered")
+        return Handle(player, format)
+    }
+
+    private inner class Handle(private val player: PlatformAudioPlayer, format: SinkFormat) : SinkHandle {
```
teancom, line 47: The handle creates a `MutableSharedFlow` with no replay, then calls Swift’s `prepareStream()` inside its constructor. If native decoder creation fails, Swift calls `listener.onError()` synchronously. That callback emits `SinkEvent.Died`.

The scheduler subscribes only after `sink.open()` returns. With no subscribers and no replay, the failure event is discarded. `extraBufferCapacity` does not preserve events for future subscribers.

Trigger: native decoder construction fails during `prepareStream()`.

Impact: the scheduler treats initialization as successful instead of delivering the intended sink-failure event. Subsequent iOS writes report success regardless of native decoding, so the pipeline can report playback without functioning audio.

Suggested fix: propagate synchronous initialization failure out of `open()`, or retain initialization status until the scheduler observes it.

Regression test: use a fake native player that calls `onError()` synchronously from `prepareStream()`, and assert that the pipeline receives and handles the failure.

Related code:  `NativeAudioController.swift:179–195`

teancom (changes requested):
Hell yeah, looking better.

composeApp/src/androidMain/kotlin/io/music_assistant/client/player/local/AudioTrackSink.kt
teancom: Last of the Astra reviews, on the fixes for the previous findings (the other two look great). I wouldn't normally even bother you about this but I'm actually trying to track down a user-reported bug in the iOS app that only triggers after continuous playback for hours (overnight, in their case), which is exactly this sort of thing.

The new `MonotonicFrameCounter` receives both position sources through one shared counter. A timestamp can represent an older position than a preceding fallback playback-head reading.

With `R = 2³²`, this sequence breaks it:

Reading                    │ Returned extended position
Timestamp: `R − 100`       │ `R − 100`
Fallback after wrap: `50`  │ `R + 50`
Older timestamp: `R − 100` │ `2R − 100` — wrong epoch
Current fallback: `100`    │ `2R + 100` — remains wrong

I reproduced those outputs using the actual compiled Kotlin class, not just a rewritten arithmetic model. The older timestamp in this example is only 150 frames behind the fallback—about 3.1 ms at 48 kHz.

This leaves the scheduler with a played-frame count far ahead of written frames, causing its queued-audio calculation to clamp to zero. The clean-wrap tests do not cover this interleaving.

Suggested correction: map older observations to the appropriate wrap epoch without advancing or corrupting the latest-position tracking. Add this exact interleaved-source regression test.

seadowg (changes requested):
Will keep testing later today and going to have at least a scan of the code. One change I'd like to see though: a `README`! 

I know that the `.claude/architecture.md` has been updated, but I generally find including a human facing `README.md` with a module is a great way to help people (and agents I imagine) understand how to interact with and think about it. This should lay out the intended use of it from a client's perspective as if it were documentation for this module as a published framework.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+The player runs for the life of `scope`. It has no `start` and no `stop`: a `null` config
+disables it, and that is the only stop.
+
+The package `io.music_assistant.sendspin.api` plus this factory are the whole public surface.
+Every other declaration in the module is `internal`, so the compiler rejects a leak into the
+application.
+
+### Configuration
+
+`LocalPlayerConfig` has two classes of field.
```
seadowg: Maybe: 

> `LocalPlayerConfig` has two classes of field: ones that will cause a reconnection when they change, and ones that don't.

And then the table could have `Causes reconnect?` as a column.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+| Field | Class | Effect of a change |
+| --- | --- | --- |
+| `endpoint` | Reconnect | Restarts the connection. The audio pipeline is untouched. |
+| `deviceName` | Reconnect | Restarts the connection. |
+| `codecPreference` | Reconnect | Restarts the connection. |
+| `userDelayMs` | Live | Applies to the next scheduled chunk. |
+| `bufferCapacityBytes` | Live | Applies to the buffer now. Advertised at the next hello. |
+
+Keep the same `Endpoint` instance while the endpoint did not change. `Endpoint.WebRtc` compares
```
seadowg: "Clients should keep the same..."

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+by identity, so a new instance restarts the connection.
+
+### Ports the application implements
+
+| Port | Purpose | Implementation |
+| --- | --- | --- |
+| `AudioSink` | Platform audio output. `open` builds a new device stream each time. | `AudioTrackSink.kt` (Android), `AudioQueueSink.kt` (iOS) |
```
seadowg: I think in general, this readme shouldn't reference implementation details in consumer modules. Those are details it doesn't need to know about.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+| --- | --- |
+| `PlaybackStarted` | Nothing. The first audio reached the sink. |
+| `PlaybackStopped(cause)` | Nothing. Show the cause if it helps the user. |
+| `ServerRefreshNeeded` | Refetch the player list. The server assigned or changed the player id. |
+| `FocusRegained` | Resume playback if the user wants it. |
+| `Warning(code)` | Show a notice. The module already recovered or degraded. |
+
+## Architecture
+
+### Package dependencies
```
seadowg: I'd maybe ditch this section. It's super hard to keep architecture documentation at this level up to date in my experience - stuff will and should be able to move around.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+- No class implements `CoroutineScope`. Cancellation is the only teardown.
+- The session reader never blocks on the sink. It must stay free to process `stream/clear`.
+- The audio thread is the single owner of the decoder and the sink handle.
+- The audio pipeline outlives the connection, so buffered audio drains through a reconnect.
+
+## Connection lifecycle
+
+```mermaid
+stateDiagram-v2
```
seadowg, line 148: I think this line is redundant?

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+## Protocol
+
+The session uses `Noise_KKpsk2_25519_ChaChaPoly_SHA256`. **The server is the Noise initiator and
+this client is the responder**, whichever side opened the connection. See
+[`noise/README.md`](src/commonMain/kotlin/io/music_assistant/sendspin/noise/README.md) for the
+cryptography, the specification mapping, and the test vectors.
+
+One attempt runs these steps:
+
+1. For a WebSocket endpoint only: send `auth` with the token and wait for `auth_ok`. This is the
```
seadowg: Why does the sendspin player have to use MA auth?

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+activation, and it closes again during a pairing activity and during a re-handshake. All sends
+take one mutex, because Noise nonces must stay in order. `ActivationPolicy.kt` holds the
+admission rules as a pure function: it admits, rejects with `pairing_required` or `unauthorized`,
+or aborts a pairing attempt with an unsupported method.
+
+The server can start a re-handshake inside the encrypted channel. The gate closes, the previous
+handshake hash becomes the new prologue, the reply goes out under the old keys, and the keys
+swap. Granted roles and any pairing attempt in progress are discarded with the old keys.
+
+## Audio path
```
seadowg: Again, I think this goes too depp. This is probably better as KDoc above the relevant classes (if they're not documented already).

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+`ClockQuality` reports `Good`, `Degraded` when the minimum round-trip time is high, and `Lost`
+when no sample was accepted for a long time. The constants live in `ClockSync.kt` and
+`ClockProbe.kt`.
+
+The probe loop is also the liveness watchdog. A server that answers no probe for three
+consecutive bursts is declared silent, which ends the attempt. This works over any transport,
+because it needs the protocol only.
+
+## Why this design
```
seadowg: Maybe have this at the top and call it "Design goals" or something? It feels more like a vision than justification. It's also potentially covered by "Scope".

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+### Ports the application implements
+
+| Port | Purpose | Implementation |
+| --- | --- | --- |
+| `AudioSink` | Platform audio output. `open` builds a new device stream each time. | `AudioTrackSink.kt` (Android), `AudioQueueSink.kt` (iOS) |
+| `DecoderFactory` | Creates one `AudioDecoder` for a codec. | `AndroidDecoders.kt`, `IosDecoderFactory` in `AudioQueueSink.kt` |
+| `SendspinKeyStore` | Byte-blob storage for identity and trust. Reads never throw. | `SettingsSendspinKeyStore.kt` |
+| `SendspinTransport` | Frames over one WebRTC data channel. | `DataChannelTransport` in `LocalPlayerEndpoints.kt` |
+
+`SendspinDeps` also takes the application's `HttpClient`, an `online` flow, a `pairWebPlayer`
```
seadowg: On the initial read, I'm finding the `pairWebPlayer` concept hard to understand. It seems like there's some interaction with the MA API (outside of the sendspin protocol), and I'm not sure I understand why that needs to happen.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+the config flow, and it creates the player. It also mirrors the server-assigned player id back
+into the settings, and implements `pairWebPlayer` over the Music Assistant API.
+
+`MainDataSource.kt` is a one-way consumer. It re-exports the state, reacts to
+`ServerRefreshNeeded`, and sends all transport commands over the Music Assistant API.
+
+Registration points: `SharedModule.kt` for the key store, the endpoints, and the adapter;
+`AndroidModule.kt` and `IosModule.kt` for the sink and the decoder factory.
+
+## Build and test
```
seadowg: Don't think this really needs documented for a gradle submodule.

sendspin/README.md
```diff
@@ -0,0 +1,355 @@
...
+## Why this design
+
+1. One owner per piece of state.
+2. One reconnect policy, in one pure object.
+3. Bounded queues everywhere. Nothing on the audio path grows without a limit.
+4. Structured concurrency. Cancellation is the only teardown.
+5. The audio pipeline outlives the connection, so buffered audio drains through a reconnect.
+6. A small public surface, enforced by `internal` and checked by the compiler.
+
+## How the application wires it
```
seadowg: Again, this is probably better as individual file KDoc if it's not there already.

seadowg (changes requested):

seadowg (changes requested):
> A Sendspin server normally needs a human to approve a pairing code. MA's own web players skip that: the server accepts a pairing token over MA's API instead. We are such a web player, but the module has no MA API client, so it calls back into the app. Fires when a session comes up unpaired, and again after a server-side unpair. Anyone talking to a Sendspin server directly can pass a no-op and pair by code.

Right. So instead of a PIN exchange, we send a token over the authenticated MA API. This makes sense, but I think the naming really through me off. Could we call it just `pairPlayer`? I think the "web" here is very distracting (even if that's what the API command uses).

Also, if we send the pairing token over `ws://`, it's still just plain text right? I'm guessing that's OK because the token is just public key material as far as I can see from [the spec](https://www.sendspin-audio.com/build/spec/#pairing-token). Also, if we're in a plain text environment, my MA auth token is exposed anyway, but just making sure I understand what's going on here 😆.

> The token is the proxy's price of entry, not Sendspin's. The default endpoint is MA's own socket at <server>/sendspin. MA puts the Sendspin server behind the same authenticated reverse proxy as the control socket, so the client needs one open port and one credential. ProxyAuth sends auth / waits for auth_ok before the Sendspin protocol starts. WebRTC skips it; the channel is already proven.

I'm having a hard time understanding this still. Is this the "normal" scenario of using the direct connection? Does MA not expose the Sendspin port? I'm just confused as to how non-authed clients (like `sendspin-cli`) then work. 

> regarding the occasional stop - didn't notice that, I think you're playing music a lot. :)
We can investigate that separately - this PR is already huge, didn't want to add logging and traces here...

Haha, yes I've been using way more than I usually would - I mostly use the desktop app as the player controller by the mobile app day to day while I'm working. I agree the current stability is in good shape for a "beta". The desktop app and CLI were both more flakey even just a couple of months ago.

seadowg (approved):
That breakdown of the different ways of connecting is brilliant and I think we should include in the README! 

I'm coming at this from just having read the pre-Noise sendspin spec, only using it locally and using other sendspin clients like sendspin-cli, so in my mind Sendspin always worked via the "Direct connection, two ports" scenario. Really helpful to understand the MA "proxy" approach.

I'd still really like to take some time to look at tests and also have a think about if we want another layer on between this and our Compose stuff that more fully encompasses playback (AudioSink implementations etc.), but I don't want to hold this up given the limited time I can give you right now if you're keen to merge and get more feedback on it. I feel like everything I'm thinking about now can be built on top of these foundations!

teancom (approved):
Heck yeah, let's do it. Do you want me to sort out the merge conflict if you're afk @formatBCE ?

formatBCE:
@teancom yes please. :) Didn't even realize there are conflicts now. Daheck :)

PR #980, opened 2026-09-08
Log once per session if we get an ATS error

teancom:

This will help diagnose if missing images are due to ATS or not without overwhelming the logs with a line per broken image.


formatBCE (approved):
Leave it with you @teancom  :)

PR #985, opened 2026-09-08
Support Android 13+ per-app language preferences

### LINKED ISSUE #984: Android language setting
I'm learning a new language, so my phone's system language is set to it, but since I'm still figuring out the MA UI, I don't want to risk misreading text. Please support Android 13+'s per-app language API ([AppCompatDelegate.setApplicationLocales](http://appcompatdelegate.setapplicationlocales/)) so I can keep MA in my native language while the rest of the phone stays in the language I'm learning. An in-app language picker under Settings -> Advanced would be a nice fallback for older Android versions.

karan:
- Declares `android:localeConfig="@xml/locales_config"` in `androidApp/src/main/AndroidManifest.xml`.
- Adds a `@CacheableTask` Gradle task `GenerateLocalesConfig` in `androidApp/build.gradle.kts` that derives `locales_config.xml` at build time directly from the `composeApp/src/commonMain/composeResources/values*` directories and wires it into `sourceSets["main"].res.srcDir`.
  - Automatically avoids translation drift as new translations are pulled from Lokalise.
  - Fails fast with a descriptive build error if an unfamiliar qualifier format is encountered, prompting explicit handling.

### Test Screenshots (Android 16 Emulator)

image

image

Closes #984

This allows users to choose their preferred language for Music Assistant directly in the Android System Settings (**Settings → Apps → Music Assistant → Language**) independently of their device's global system language.
Because `MainActivity` already declares `locale|layoutDirection` in `configChanges`, Compose Multiplatform automatically re-evaluates localized strings via `LocalConfiguration.current` when the language changes without restarting the app.

N/A


formatBCE (commented):
Delete androidApp/src/main/res/xml/locales_config.xml. Keep the manifest attribute. Add to androidApp/build.gradle.kts:
```
@CacheableTask
abstract class GenerateLocalesConfig : DefaultTask() {
    @get:InputDirectory
    @get:PathSensitive(PathSensitivity.RELATIVE)
    abstract val resourcesDir: DirectoryProperty

    @get:Input
    abstract val baseLanguage: Property<String>

    @get:OutputDirectory
    abstract val outputDir: DirectoryProperty

    @TaskAction
    fun generate() {
        val qualifier = Regex("""^values-([a-z]{2,3})(?:-r([A-Z]{2}))?$""")
        val locales = resourcesDir.get().asFile.listFiles().orEmpty()
            .filter { it.isDirectory && it.name.startsWith("values") }
            .map { dir ->
                if (dir.name == "values") baseLanguage.get()
                else qualifier.matchEntire(dir.name)
                    ?.groupValues?.drop(1)?.filter(String::isNotEmpty)?.joinToString("-")
                    ?: error("Unsupported locale directory '${dir.name}'. Teach GenerateLocalesConfig its BCP-47 form.")
            }
            .sorted()

        outputDir.get().asFile.resolve("xml/locales_config.xml").apply { parentFile.mkdirs() }
            .writeText(
                buildString {
                    appendLine("""<?xml version="1.0" encoding="utf-8"?>""")
                    appendLine("""<locale-config xmlns:android="http://schemas.android.com/apk/res/android">""")
                    locales.forEach { appendLine("""    <locale android:name="$it" />""") }
                    append("</locale-config>")
                }
            )
    }
}

// Android needs an explicit locale list for the per-app language picker. The shipped
// languages are the Lokalise-synced `values-*` directories, so the list is derived from
// them; a hand-written copy drifts on the next translation pull without any build error.
val generateLocalesConfig by tasks.registering(GenerateLocalesConfig::class) {
    resourcesDir.set(rootProject.layout.projectDirectory.dir("composeApp/src/commonMain/composeResources"))
    baseLanguage.set("en")
    outputDir.set(layout.buildDirectory.dir("generated/res/localesConfig"))
}

android {
    sourceSets["main"].res.srcDir(generateLocalesConfig.flatMap { it.outputDir })
}
```

Two design points:
- error(...) on an unknown qualifier, not a silent skip. If CMP ever gains script subtags and a values-b+sr+Latn appears, the build must stop and make someone decide. Dropping the locale quietly is exactly the failure the manual list already has.
- Input is the directory, so a translation content change also reruns the task. Slightly over-eager, but it runs weekly at most and costs microseconds. Declaring only the names as an input would need config-time directory listing, which is worse for the configuration cache. Not worth the cleverness.

formatBCE (approved):
Great job! Thank you for contribution!

PR #994, opened 2026-09-11
Favourite the song on air from a radio stream (Android Auto + player)

### LINKED ISSUE #992: Android Auto: no way to favourite the track playing on a radio stream
## Problem

When a radio stream is playing, there is no way to favourite the *song* that is currently on air from the Android Auto (or notification) media session. The favourite action simply isn't offered.

The action itself already exists — `ACTION_TOGGLE_FAVORITE` is registered in `SharedMediaSessionManager` — but it is gated to library tracks:

```kotlin
// androidApp/.../services/MediaNotificationData.kt:75-76
isFavoritableTrack = currentTrack
    ?.let { it.mediaType == MediaType.TRACK && it.canBeFavorited } == true,
```

consumed by `MediaSessionActions.kt:54` (`SessionAction.FAVORITE -> isFavoritableTrack`). A radio item is `MediaType.RADIO`, so the slot is never filled and the button never reaches the car screen.

Loosening that flag alone wouldn't be correct, though. The handler routes to `MainDataSource.toggleFavorite`, which calls `Request.Library.addFavorite(item.uri)` — for a radio item that URI is the **station**, so it would favourite the station rather than the track on air. That is a different (and much less useful) thing than what a listener means when they hit ♥ during a radio show.

This matters most in the car, where Android Auto is the only surface available and hearing something good on a radio stream is exactly when you want to capture it.

## Solution

The server already has the right primitive: `players/add_currently_playing_to_favorites` (arg `player_id`, available from schema version 27). It resolves the currently playing stream's metadata — the ICY `Artist - Title` — to an actual library media item and favourites that. Nothing in the app calls it today.

Proposed:

1. Add the command to `Request` (e.g. `Request.Players.addCurrentlyPlayingToFavorites(playerId)`).
2. Offer the FAVORITE session action when the current item is `MediaType.RADIO` and the player is playing, in addition to the existing favouritable-track case.
3. In the `ACTION_TOGGLE_FAVORITE` handler, branch on media type: library tracks keep today's `toggleFavorite` behaviour; radio calls the new player command instead.
4. Treat the radio case as add-only rather than a toggle — the queue payload reports the *station's* favourite flag, not the resolved track's, so there is no meaningful "unheart" state to render. The server raises if the stream carries no title or the track can't be resolved, which should surface as a no-op/error rather than an optimistic flip.

Happy to work on this if the approach sounds right — particularly on point 4, since there may be a preference about whether the action should be hidden or shown-and-failing when a stream has no usable ICY metadata (plenty don't).

Environment: Music Assistant server 2.10.0 (schema 68), app 0.13.0 on Android 17.

tannernicol:
Closes #992

**What it does.** When a radio stream has a real song on air, the favourite action is now offered — in the Android Auto / notification media session and on the in-app player heart — and it calls the server's `players/add_currently_playing_to_favorites` for that player. The server resolves the stream's current title to a library item and favourites *that*, so the listener hearts the song, not the station.

**The "song on air" signal is the one the app already uses to display one.** The inline rule in `NowPlayingChannels.withRadioStreamMetadata` (RADIO item, `currentMedia.queueItemId == currentItem.id`, non-blank title that isn't the station name) is extracted into `radioStreamTitle` / `PlayerData.hasFavoritableStreamTrack()` and shared by Now Playing, the media session and the player UI, so all three agree for any server and provider. No stream-specific metadata is assumed.

**Add-only, un-filled heart.** The queue payload's `favorite` flag belongs to the station, not the on-air song, so there is no truthful "already favourited" state for a stream: the stream case never toggles and always renders the outline heart. Repeated taps are idempotent server-side. Library-track favouriting is untouched — the TRACK branch in `ACTION_TOGGLE_FAVORITE` and `toggleFavorite` are as before.

**Version gate.** The command exists from server schema 27; `supportsFavoriteCurrentlyPlaying(schemaVersion)` follows the existing `EncryptionModeGate` / `LeaderLeaveSupport` pattern and treats an unknown schema as unsupported.

**Unsupported servers and expected refusals stay silent.** The stream ♥ is offered only when there is a song on air *and* the connected server has `players/add_currently_playing_to_favorites` (schema ≥ 27, gated the same way as `EncryptionModeGate` / `LeaderLeaveSupport`); on an older server nothing is shown and nothing is logged, so behaviour is identical to today. When the server can't resolve the title (no ICY metadata, or the song isn't in any library) that's treated as an expected refusal — no error log, no UI — matching how `toggleFavorite`'s failure is handled.

None. The `withRadioStreamMetadata` refactor is behaviour-preserving (the existing `NowPlayingRadioStreamMetadataTest` cases still pass unchanged); no translations, generated files or dependencies touched.

- [x] Made sure checks pass by running `./gradlew detektAll testAndroidHostTest :androidApp:testDebug` — BUILD SUCCESSFUL, 806 tests (791 before), 0 failures, detekt clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0198cQrZ4nN32gxCGpwtZy9Q

### DIFF
--- a/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaNotificationData.kt
+++ b/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaNotificationData.kt
@@ -1,6 +1,7 @@
 package io.music_assistant.client.services
 
 import android.os.SystemClock
+import io.music_assistant.client.data.hasFavoritableStreamTrack
 import io.music_assistant.client.data.model.client.MediaType
 import io.music_assistant.client.data.model.client.PlayerData
 import io.music_assistant.client.data.model.client.RepeatMode
@@ -27,6 +28,9 @@ data class MediaNotificationData(
     // Current item is a favoritable track: a favorite toggle competes for a slot
     // (see sessionActions for the slot rule).
     val isFavoritableTrack: Boolean,
+    // Current item is a radio stream with a real song on air: same slot competition
+    // as isFavoritableTrack, but the action always adds (see getFavoriteIcon).
+    val isFavoritableStream: Boolean,
     val isFavorite: Boolean,
     val isPlaying: Boolean,
     val imageUrl: String?,
@@ -61,6 +65,7 @@ data class MediaNotificationData(
             currentChapter: ResolvedChapter? = null,
         ) = run {
             val currentTrack = playerData.queueInfo?.currentItem?.track as? AppMediaItem
+            val isFavoritableStream = playerData.hasFavoritableStreamTrack()
             MediaNotificationData(
             multiplePlayers = multiplePlayers,
             longItemId = playerData.player.currentMedia?.hashCode()?.toLong(),
@@ -74,7 +79,11 @@ data class MediaNotificationData(
             isLongFormContent = playerData.queueInfo?.currentItem?.track.isLongFormSpokenContent,
             isFavoritableTrack = currentTrack
                 ?.let { it.mediaType == MediaType.TRACK && it.canBeFavorited } == true,
-            isFavorite = currentTrack?.favorite == true,
+            isFavoritableStream = isFavoritableStream,
+            // The station's own favorite flag would show a filled heart for an already-
+            // favorited station even though nothing has been favorited for the on-air song,
+            // so the stream case always renders un-filled.
+            isFavorite = currentTrack?.favorite == true && !isFavoritableStream,
             isPlaying = playerData.player.isPlaying,
             imageUrl = playerData.player.currentMedia?.imageUrl,
             chapterName = currentChapter?.displayName,
--- a/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaNotificationData.kt
+++ b/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaNotificationData.kt
@@ -1,7 +1,6 @@
 package io.music_assistant.client.services
 
 import android.os.SystemClock
-import io.music_assistant.client.data.hasFavoritableStreamTrack
 import io.music_assistant.client.data.model.client.MediaType
 import io.music_assistant.client.data.model.client.PlayerData
 import io.music_assistant.client.data.model.client.RepeatMode
@@ -63,9 +62,11 @@ data class MediaNotificationData(
             multiplePlayers: Boolean,
             effectiveElapsedSec: Double?,
             currentChapter: ResolvedChapter? = null,
+            // Real on-air stream song AND the connected server can resolve it — see
+            // MainDataSource.canFavoriteCurrentlyPlaying, the single source for this.
+            isFavoritableStream: Boolean = false,
         ) = run {
             val currentTrack = playerData.queueInfo?.currentItem?.track as? AppMediaItem
-            val isFavoritableStream = playerData.hasFavoritableStreamTrack()
             MediaNotificationData(
             multiplePlayers = multiplePlayers,
             longItemId = playerData.player.currentMedia?.hashCode()?.toLong(),
--- a/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaSessionActions.kt
+++ b/androidApp/src/main/kotlin/io/music_assistant/client/services/MediaSessionActions.kt
@@ -51,7 +51,7 @@ internal fun sessionActions(data: MediaNotificationData): List<SessionAction> = 
 }.take(SLOT_COUNT)
 
 private fun MediaNotificationData.supports(action: SessionAction) = when (action) {
-    SessionAction.FAVORITE -> isFavoritableTrack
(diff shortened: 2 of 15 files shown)


formatBCE (commented):
Thank you, that's a solid PR. It fits the conventions of this repo well.

Found four things. The first is blocking.

---

### 1. The in-app heart never reaches the new branch (blocking)

`PlayerItems.kt:687` gates the first branch on `currentTrack?.canBeFavorited == true`, and `AppMediaItem.kt:131` defines `canBeFavorited` as `uri != null`. A radio station always carries a URI, so the station branch always wins and the new `else if (canFavoriteStream)` branch is unreachable.

The notification path works only because `MediaNotificationData.from` adds a second condition — `mediaType == MediaType.TRACK`. The in-app player has no equivalent.

The net effect is a split gesture: the notification favourites the song, the in-app heart favourites the station, same icon, no indication which. Please reorder so an on-air song outranks the station:

```kotlin
if (canFavoriteStream) {
    // Radio with a song on air: the heart adds the song, not the station.
    ...
} else if (currentTrack?.canBeFavorited == true) {
    ...
}
```

Please also add a test pinning the branch order for a radio station that has a URI. No current test covers it, which is how this got through.

### 2. The gesture gives no feedback at all

`toggleFavorite` flips an optimistic override so the track heart fills immediately. `favoriteCurrentlyPlaying` correctly has no override — the server resolves the item from the stream title, so there is nothing local to flip — but nothing replaces it. The user taps the heart and nothing changes, on success or on failure. It will read as a broken button, and the server's "cannot resolve the title" refusal is common rather than exceptional.

`ActionsViewModel` already owns the mechanism: `_toasts`, used by `onLibraryClick`, the playlist actions and the mark-played actions (`ActionsViewModel.kt:82-164`).

Suggested shape:

1. Make `MainDataSource.favoriteCurrentlyPlaying` a `suspend` function returning the `Result` from `sendRequest`, instead of a fire-and-forget `launch`. Keep the `canFavoriteCurrentlyPlaying` guard and early return.
2. In `SharedMediaSessionManager`, call it inside the manager's own scope.
3. In `ActionsViewModel.onFavoriteStreamClick`, launch in `viewModelScope` and emit `onSuccess`/`onFailure` into `_toasts`, exactly like `markPlayed` at `ActionsViewModel.kt:145-147`.
4. Add the two strings to `composeApp/src/commonMain/composeResources/values/strings.xml` only — Lokalise syncs the other locales, do not hand-add translations. Suggested keys: `toast_favorited_stream_track`, `toast_error_favorite_stream_track`. Make the failure string name the cause so it does not read as a crash.

### 3. The schema gate is read imperatively during composition

`PlayersPager.kt:339` calls `actionsViewModel.canFavoriteStream(player)` during composition, and that reads `apiClient.sessionState.value`. That is not a snapshot read, so composition does not observe it — the flag only refreshes when something else happens to recompose the pager. `HomeScreenViewModel.kt:100-112` sets the pattern here.

The two halves of the predicate have different lifetimes, so split them:

- Server half (session-scoped) → a `StateFlow<Boolean>` on the view model: `apiClient.sessionState.map { supportsFavoriteCurrentlyPlaying((it as? HasConnectionData)?.serverInfo?.schemaVersion) }.stateIn(...)`.
- Content half (per player, already pure) → call `player.hasFavoritableStreamTrack()` at the call site, where `PlayerData` already drives recomposition.
- Combine at the call site: `canFavoriteStream = streamFavoriteSupported && player.hasFavoritableStreamTrack()`.

Keep `MainDataSource.canFavoriteCurrentlyPlaying` as the guard inside `favoriteCurrentlyPlaying`. An imperative read is the right thing there.

### 4. Placement (suggestion, not blocking)

`hasFavoritableStreamTrack` and `canFavoriteCurrentlyPlaying` are public extensions declared in `NowPlayingChannels.kt`, which is the iOS/CarPlay now-playing bridge. They are domain predicates, not bridge data; they live there only because they share the private `radioStreamTitle` helper. Consider moving `radioStreamTitle` and both predicates next to the other favourite support code and importing them back into `NowPlayingChannels.kt`. File move only, no logic change.

---

### Please do not change these — they are correct

- The schema floor of 27, and its reasoning.
- The absence of an optimistic override.
- `isFavorite = currentTrack?.favorite == true && !isFavoritableStream`. The station's own favourite flag would otherwise draw a filled heart for a song nobody favourited.
- The `queueItemId` and "title equals station name" guards in `radioStreamTitle`.

### What I would like to see before merge

1. `./gradlew :composeApp:compileCommonMainKotlinMetadata` and `./gradlew :androidApp:compileDebugKotlin` pass.
2. `./gradlew :composeApp:jvmTest :androidApp:testDebugUnitTest` pass, including the new branch-order test.
3. On a device: a radio station reporting a song title — the in-app heart adds the song and shows the success toast, and the notification heart does the same.
4. A radio station with no song title — the heart falls back to the station gesture.
5. If you have one: a server below schema 27 — the stream heart does not show.

PR #995, opened 2026-09-11
Match artists within the same provider

seadowg:
Work towards #801

This tidies up artist pages where an artist has multiple mappings across a single provider (like when they have collaboration albums for example).

Last thing to follow-up with is showing provider names instead of the raw domain.

The main change to get this working was to expand the `ItemList` concept so it can support multiple mappings allowing to easily list items from across them. That

### DIFF
--- a/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/FetchArtistItemsUseCase.kt
+++ b/composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/FetchArtistItemsUseCase.kt
@@ -3,28 +3,26 @@ package io.music_assistant.client.data.model.client.items
 import io.music_assistant.client.data.model.server.ProviderMapping
 import io.music_assistant.client.data.repository.MediaItemRepository
 import io.music_assistant.client.ui.compose.item.ItemList
-import io.music_assistant.client.ui.compose.item.toRequest
+import io.music_assistant.client.ui.compose.item.toRequests
 
 class FetchArtistItemsUseCase(private val mediaItemRepository: MediaItemRepository) {
     suspend fun run(
         artist: Artist,
-        itemListBuilder: (ProviderMapping) -> ItemList,
+        itemListBuilder: (List<ProviderMapping>) -> ItemList,
     ): ArtistItems? {
         if (artist.providerMappings.isNullOrEmpty()) {
             return null
         }
 
-        val itemLists = artist.providerMappings.map { itemListBuilder(it) }
         val providers = artist.providerMappings.groupBy { it.providerInstance }.map { it.value }
-        for (mappings in providers) {
-            val items = mappings.flatMap {
-                val itemList = itemListBuilder(it)
-                val result = mediaItemRepository.fetchMediaItems(itemList.toRequest())
+        val itemLists = providers.map { itemListBuilder(it) }
+        for (itemList in itemLists) {
+            val items = itemList.toRequests().flatMap {
+                val result = mediaItemRepository.fetchMediaItems(it)
                 result.getOrNull() ?: emptyList()
             }
 
             if (items.isNotEmpty()) {
-                val itemList = itemListBuilder(mappings.first())
                 return ArtistItems(items, itemList, itemLists)
             }
         }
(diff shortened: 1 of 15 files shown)


composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/FetchArtistItemsUseCase.kt
formatBCE, line 1: ~~Weird formatting~~
Never mind, GitHub is broken I guess...

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
```diff
@@ -95,17 +96,25 @@ class MediaItemDataMediator(
...
+                val items = mediaItemRepository.fetchMediaItems(it).getOrEmptyList()
+                stateFlow.value = DataState.Data(items)
             } catch (_: Exception) {
                 stateFlow.value = DataState.Error()
             }
         }
     }
 }
 
+suspend fun MediaItemDataMediator.set(request: Request) {
```
formatBCE, line 111: Why is this extension function? It's perfect class function, though I guess deprecated...

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
```diff
@@ -95,17 +96,25 @@ class MediaItemDataMediator(
...
+    this.set(listOf(request))
+}
+
+fun MediaItemDataMediator.set(items: List<AppMediaItem>, request: Request) {
+    this.set(items, listOf(request))
```
formatBCE, line 116: same

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
```diff
@@ -194,5 +195,23 @@ suspend fun MediaItemRepository.fetchRecommendationFolders(): Result<List<Recomm
     )
 }
 
+/**
+ * Fetch items for multiple requests and combine them into one list. Returns a failure Result
```
formatBCE, line 199: I think we will have questions to this behaviour later. :)

composeApp/src/commonMain/kotlin/io/music_assistant/client/ui/compose/item/ItemList.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.music_assistant.client.data.model.server.ServerMediaItem
+import kotlinx.serialization.Serializable
+
+@Serializable
+sealed interface ItemList {
+    val mediaType: MediaType
+    val providerDomain: String
+
+    @Serializable
+    data class ArtistAlbums(val mappings: List<ProviderMapping>) : ItemList {
```
formatBCE: `providerMappings`? To conform to single mapping constructor.

formatBCE (approved):
Minor comments, feel free to merge as-is

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
seadowg: It's definitely a pattern I've started adopting more and more (in languages with extensions): if a function doesn't rely on internal state, then it can just be an extension rather than an instance one. Keeps the actual interface surface as small as possible and avoids the risk of disparate implementations in overloads.

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
seadowg: Same as https://github.com/music-assistant/mobile-app/pull/995#discussion_r3990946213

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/repository/MediaItemRepository.kt
seadowg: Yeah I think error handling and performance are both things that could be tweaked/improved. Hard to make a call about either for these simple cases though (always a single provider, uncommon scenario and so on).

composeApp/src/commonMain/kotlin/io/music_assistant/client/data/model/client/items/MediaItemDataMediator.kt
formatBCE: Makes sense!

formatBCE (approved):

PR #1002, opened 2026-09-12
Support dynamic text resizing in iOS

teancom:

While working on #1001 I noticed that we didn't respond to the default text size changing in iOS dynamically. The user would have to completely close and re-open the app. We now scale up and down "live" (it's iOS so you're switching back and forth between the Settings app and MA).

Also bump our compose-multiplatform version because 1.11.1 had a bug that prevented dynamic text resizing from working, which was fixed in 1.12.0.

### DIFF
--- a/composeApp/src/iosMain/kotlin/io/music_assistant/client/ui/compose/DynamicTypeFontScale.ios.kt
+++ b/composeApp/src/iosMain/kotlin/io/music_assistant/client/ui/compose/DynamicTypeFontScale.ios.kt
@@ -1,0 +1,76 @@
+@file:OptIn(kotlinx.cinterop.ExperimentalForeignApi::class)
+
+package io.music_assistant.client.ui.compose
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.unit.Density
+import platform.Foundation.NSNotificationCenter
+import platform.Foundation.NSOperationQueue
+import platform.UIKit.UIApplication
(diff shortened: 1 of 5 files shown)


composeApp/src/iosMain/kotlin/io/music_assistant/client/ui/compose/DynamicTypeFontScale.ios.kt
```diff
@@ -0,0 +1,76 @@
...
+import platform.UIKit.UIContentSizeCategoryLarge
+import platform.UIKit.UIContentSizeCategoryMedium
+import platform.UIKit.UIContentSizeCategorySmall
+
+@Composable
+actual fun ProvideDynamicTypeFontScale(content: @Composable () -> Unit) {
+    var fontScale by remember { mutableStateOf(currentFontScale()) }
+
+    DisposableEffect(Unit) {
+        val observer = NSNotificationCenter.defaultCenter.addObserverForName(
```
seadowg, line 36: Just one thought here: it makes sense to observe and clean-up with `DisposableEffect` here, but I'm wondering if we should also be using the local lifecycle to manage the observation to prevent it being active while the app is in the background? Very usefully, the docs for `DisposableEffect` use that setup as an example: https://developer.android.com/develop/ui/compose/side-effects#disposableeffect.

It could be that that doesn't make sense for iOS (and we are in platform specific code here).

seadowg (changes requested):

composeApp/src/iosMain/kotlin/io/music_assistant/client/ui/compose/DynamicTypeFontScale.ios.kt
teancom: Good question! The local lifecycle is available on iOS too, so we could do that. In this case, though, this is just a passive notification observer rather than something doing ongoing background work, and `onDispose` already handles cleanup. And since we'd be in the background when the user is changing the font size (since that's in the Settings app), stopping observation just means we have to re-read the scope when we start observing again. I’m inclined to keep this composition-scoped unless there’s a particular background behavior we need to avoid. Does that make sense?

(And once Compose fixes their iOS-specific bug it can all be dropped anyway)

seadowg (approved):

PR #1003, opened 2026-09-12
Limit CarPlay radio controls to play and pause

teancom:

I was listening to an internet radio station in the car and noticed it had prev/next/shuffle/repeat all enabled as controls, none of which were actually useful. This detects 'radio mode' and disables everythign except play/pause.

### DIFF
--- a/composeApp/src/commonTest/kotlin/io/music_assistant/client/data/NowPlayingChannelsTest.kt
+++ b/composeApp/src/commonTest/kotlin/io/music_assistant/client/data/NowPlayingChannelsTest.kt
@@ -145,6 +145,25 @@ class NowPlayingRadioStreamMetadataTest {
         )!!
 
     @Test
+    fun radioClassificationSurvivesStreamMetadataChanges() {
+        assertTrue(radioTrack(null).isRadio)
+        assertTrue(radioTrack(streamMedia()).isRadio)
+        assertTrue(radioTrack(streamMedia(title = "Another song")).isRadio)
+        assertFalse(radioTrack(streamMedia()).isLongFormContent)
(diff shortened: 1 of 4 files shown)


composeApp/src/commonTest/kotlin/io/music_assistant/client/data/NowPlayingChannelsTest.kt
```diff
@@ -144,6 +144,25 @@ class NowPlayingRadioStreamMetadataTest {
...
+        assertTrue(radioTrack(streamMedia(title = "Another song")).isRadio)
+        assertFalse(radioTrack(streamMedia()).isLongFormContent)
+        for (item in listOf(testTrack(), testAudiobook(), testPodcastEpisode())) {
+            assertFalse(buildNowPlayingTrack(playerData(item, queueInfo(queueId = "queue-1")))!!.isRadio)
+        }
+    }
+
+    @Test
+    fun radioDisablesQueueModeControls() {
+        for (dynamic in listOf(false, true)) {
```
seadowg: Personally, using a `for` here feels harder to read than just unrolling to 2 setup/assertion pairs. Plus, using loops in a test like this will make it harder to read failure results.

seadowg (changes requested):

teancom:
@seadowg How does this look now?

seadowg (approved):

seadowg:
@teancom looks good! Just a tip (in case you didn't just forget): you can re-request review once you've got changes made up at the "Reviewers" and it will ping me as well.

teancom:
Ah, I didn't know. Thank you!

PR #1019, opened 2026-09-18
Add Sendspin playback sync offset setting

chicco-carone:
This PR adds support for the local sendspin player in the mass app to set a static offset to correct any delay with the other players. In the settings there is the new offset slider which also supports pressing the number to input more precisely the offset via keyboard.


seadowg:
@chicco-carone I'm guessing you didn't read out [contributing guide](https://github.com/music-assistant/mobile-app?tab=contributing-ov-file): we ask that folks add an issue for discussion before submitting a PR for changes that don't already have an open issue.

@formatBCE last time we talked about this, we'd decided not to implement offsets as we didn't want the local player to have to support group sync features. Is that right?

chicco-carone:
Sorry I didn't read it carefully enough and thought that it didn't require an issue. For the global offset I don't get why it shouldn't support it. Since its in the settings the majority of the users that will be using it to play stuff won't even notice it or touch it but for the users that could need it would be useful. I was trying to move my music setup from snapcast to sendspin and I'm using a bunch of android phones as recievers but I couldn't sync them together correctly.

seadowg:
> For the global offset I don't get why it shouldn't support it. Since its in the settings the majority of the users that will be using it to play stuff won't even notice it or touch it but for the users that could need it would be useful.

Every feature requires maintenance in the future, so even if things are purely additive they should still be discussed.

> I was trying to move my music setup from snapcast to sendspin and I'm using a bunch of android phones as recievers but I couldn't sync them together correctly.

We recommend using our friend https://github.com/chrisuthe/SendSpinDroid for using Android phones as dedicated receivers - the app is designed for that exact purpose. This app is intended to provide control and local/personal playback.

formatBCE:
Yes, @seadowg is right. We don't plan to promote this app as stationary Sendspin player. There are Sendspin players for Android.

chicco-carone:
I was aware of the other sendspin player but its implementation works way worse than the one in the mass application that's why I wanted this added

seadowg:
> I was aware of the other sendspin player but its implementation works way worse than the one in the mass application that's why I wanted this added

I think the best way of forward given your use case would be to file issues or contribute there (https://github.com/chrisuthe/SendSpinDroid) instead of here. 

I'm going to close this for now - we definitely appreciate folks contributing, but this isn't a feature we're looking to include and maintain at the moment. You're welcome to file an issue if you'd like to discuss further though!

